PR #5, opened 2016-04-21
Fix Motor version, let it pull in PyMongo

ajdavis:
Motor already depends on a specific PyMongo version, better to depend on Motor instead of adding a separate dependency link to PyMongo.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,12 +16,11 @@ with open('HISTORY.rst', 'rb') as history_file:
 
 requirements = [
     "marshmallow>=2.6.0",
-    "pymongo"  # needed for bson module
 ]
 
 test_requirements = [
     "txmongo",
-    "motor>=0.7.4"
+    "motor>=0.6.2"
 ]
 
 setup(


touilleMan:
Thanks for the PR ;-)

The thing is given umongo needs the bson module (which is included in pymongo) it makes a bit clunky to count on dependencies to pull this dependency.

For example mongomock doesn't have pymongo among it dependencies so installing only umongo & monogmock would lead to a crash...

Do you think there is a way to install bson module alone ?

Another question: Why did you change motor version requirement (from 0.7.4 to 0.6.2) ?

ajdavis:
I didn't realize Motor was an *optional* dependency. In that case, including a PyMongo dependency makes sense. Be aware that Motor requires PyMongo 2.8.0 exactly, though; if umongo uses Motor it should require PyMongo 2.8.0.

No, there's no way to install bson without pymongo.

I moved the Motor version requirement to 0.6.2 because that is the latest version of Motor.

touilleMan:
I use bumpversion, but it seems to violent and change everything that have the same version of the app... so motor's version has been wrongly modified since umongo hit 0.6.2 !

PR #15, opened 2016-07-01
Add schema arg to EmbeddedDocumentImplementation dump method

lafrech:
Any reason not to have this?


touilleMan:
Can you add a test for this ? ;-)

lafrech:
I'm not sure which test you wanted me to add. The coverage is already total, here, as this is a Schema feature, tested on Marshmallow's side. Anyway, I added a test for EmbeddedDocument and a test for Document, as I don't think this was tested already. Tests pass and PR is rebased on latest master.

touilleMan:
>  Anyway, I added a test for EmbeddedDocument and a test for Document, as I don't think this was tested already

That was the tests I wanted, thanks ^^

PR #21, opened 2016-07-11
Make ListField orderable

lafrech:
Just a proof of concept. Still lacks tests/docs and perhaps comments.

Contrary to what is done in MongoEngine, I didn't create a `SortedListField` class, but added an optional sort feature to `ListField`.

### DIFF
--- a/umongo/fields.py
+++ b/umongo/fields.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+from operator import itemgetter
 from marshmallow import ValidationError, missing
 from marshmallow import fields as ma_fields
 from bson import DBRef, ObjectId, errors as bson_errors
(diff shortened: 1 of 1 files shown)


lafrech:
One thing I don't like about this implementation is that the sorting is done at commit time. So

- I can't rely on the order until the object is committed, 
- the commit modifies the object.

Not so great. I copied this from MongoEngine without thinking much...

Maybe I'll improve this in the future.

lafrech:
Another approach would be to use a [SortedListWithKey](http://www.grantjenks.com/docs/sortedcontainers/introduction.html) as underlying object. This way, the list would be kept sorted all the time.

lafrech:
I'm closing this as I'm not working on it anymore. I'll reopen if I go down to it again.

Another issue with this implementation is that it breaks if `allow_none` is `True`. The fix for this might depend on the type of the key field. I don't know how to address this in a generic fashion.

PR #22, opened 2016-07-11
Add pre_commit and post_commit hooks

lafrech:
This draft illustrates the hook feature we've been talking about by email.

It works in my simple use case. There is one limitation, though: I don't know how create a `pre_commit` method in an inheritable class and override it in a child class calling `super()` to get to the common part. I get an error:

    TypeError: super(type, obj): obj must be an instance or subtype of type

As you said, this is most likely due to the Template/Implementation duality: the methods are in `DocumentImplementation` but then I override them in `DocumentTemplate`.

I'd like to rework this to a better design, but I'd need a hint, as I don't know how to sort this out.

Did you have something else in mind?

### DIFF
--- a/umongo/document.py
+++ b/umongo/document.py
@@ -228,11 +228,11 @@ class DocumentImplementation(Implementation, metaclass=MetaDocumentImplementatio
 
     def commit(self, io_validate_all=False, conditions=None):
 
-        self.pre_commit(io_validate_all=False, conditions=None)
+        self.pre_commit(io_validate_all=False, conditions=conditions)
 
-        self._commit(io_validate_all=True, conditions=None)
+        self._commit(io_validate_all=True, conditions=conditions)
 
-        self.post_commit(io_validate_all=False, conditions=None)
+        self.post_commit(io_validate_all=False, conditions=conditions)
 
     def pre_commit(self, io_validate_all=False, conditions=None):
         pass
--- a/umongo/document.py
+++ b/umongo/document.py
@@ -226,6 +226,20 @@ class DocumentImplementation(Implementation, metaclass=MetaDocumentImplementatio
         """
         self._data.clear_modified()
 
+    def commit(self, io_validate_all=False, conditions=None):
+
+        self.pre_commit(io_validate_all=False, conditions=None)
+
+        self._commit(io_validate_all=True, conditions=None)
+
+        self.post_commit(io_validate_all=False, conditions=None)
+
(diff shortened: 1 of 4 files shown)


lafrech:
I only tested this on `pymongo` in my environment. It fails for `motor` and `txmongo`, due to the decorators on their `commit` method, I guess.

touilleMan:
I've worked on this, now there is nice `pre_update`, `post_delete` etc. methods  (see f0f51a6c251c995a665515f1698ad2e9b73764a5)

lafrech:
Great !

One question. Assume, for example's sake, you have a `date_modified` field you want to update each time you commit the document. Do you need to subclass both `pre_update` and `pre_insert` methods like this?

```py
           def pre_insert(self, payload):
               self.date_created = datetime.now()
               self.date_modified = datetime.now()

           def pre_update(self, query, payload):
               self.date_modified = datetime.now()
```

touilleMan:
Yep, that's it ! :+1:

touilleMan:
You're right about this, we should move `pre_update` and `pre_insert` to before the creation of the payload.

Providing the possibility of modifying the payload before inserting it is less useful, so we can drop this until someone ask for it.

Do you agree ?

lafrech:
Absolutely.

In practice, I think I'd move payload creation after pre_* rather than the other way around:

```py
        self.io_validate(validate_all=io_validate_all)
        # payload = self._data.to_mongo(update=self.is_created)
        try:
            if self.is_created:
                # if payload:
                # That's a use case for is_modified()
                # https://github.com/Scille/umongo/pull/41/files#diff-a39e706a68bdfe46277c7cc6cc32890fR179
                if self.is_modified():
                    query = conditions or {}
                    query['_id'] = self._data.get_by_mongo_name('_id')
                    # Likewise, do we need to pass query?
                    self.pre_update(query)
                    # Create payload here
                    payload = self._data.to_mongo(update=True)
                    ret = self.collection.update_one(query, payload)
                    if ret.matched_count != 1:
                        raise UpdateError(ret.raw_result)
                    self.post_update(ret)
                else:
                    ret = None
           elif conditions:
                raise RuntimeError('Document must already exist in database to use `conditions`.')
            else:
                self.pre_insert()
                # Create payload here
                payload = self._data.to_mongo(update=False)
                ret = self.collection.insert_one(payload)
                # TODO: check ret ?
                self._data.set_by_mongo_name('_id', ret.inserted_id)
                self.is_created = True
                self.post_insert(ret)
```

PR #24, opened 2016-07-12
Point field

lafrech:
PointField proposal.

MongoDB stores geo data conforming to GeoJSON standard (see http://blog.mongodb.org/post/50984169045/new-geo-features-in-mongodb-24), which is inplemented in [python-geojson](https://github.com/frewsxcv/python-geojson), so we don't have to recreate the [|de]serializers.

Here again, just a draft. If you like it, I may add tests/docs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,8 @@ with open('HISTORY.rst', 'rb') as history_file:
     history = history_file.read().decode('utf8')
 
 requirements = [
-    "marshmallow>=2.6.0"
+    "marshmallow>=2.6.0",
+    "geojson>=1.3.1",
 ]
 
 setup(
--- a/umongo/fields.py
+++ b/umongo/fields.py
@@ -1,4 +1,7 @@
 from datetime import datetime
+
+import geojson
+
 from marshmallow import ValidationError, missing
 from marshmallow import fields as ma_fields
 from bson import DBRef, ObjectId, errors as bson_errors
@@ -215,6 +218,28 @@ class ObjectIdField(BaseField, ma_fields.Field):
             return ObjectId(value)
         except bson_errors.InvalidId:
             raise ValidationError(_('Invalid ObjectId.'))
+
+
+class PointField(BaseField, ma_fields.Field):
+
+    def _serialize(self, value, attr, obj):
+        if value is None:
+            return None
+        return value.coordinates
+
+    def _deserialize(self, value, attr, data):
+        if isinstance(value, geojson.Point):
+            return value
+        try:
+            return geojson.Point(value)
+        except ValueError:
+            raise ValidationError(_('Invalid Point.'))
+
+    def _serialize_to_mongo(self, obj):
+        return geojson.dumps(obj)
+
+    def _deserialize_from_mongo(self, value):
+        return geojson.loads(value)
 
 
 class ReferenceField(ObjectIdField):
--- a/umongo/fields.py
+++ b/umongo/fields.py
@@ -239,7 +239,7 @@ class PointField(BaseField, ma_fields.Field):
         return geojson.dumps(obj)
 
     def _deserialize_from_mongo(self, value):
-        return geojson.loads(value)
+        return geojson.GeoJSON.to_instance(value)
 
 
 class ReferenceField(ObjectIdField):
--- a/umongo/fields.py
+++ b/umongo/fields.py
@@ -235,9 +235,6 @@ class PointField(BaseField, ma_fields.Field):
         except ValueError:
             raise ValidationError(_('Invalid Point.'))
 
-    def _serialize_to_mongo(self, obj):
-        return geojson.dumps(obj)
-
     def _deserialize_from_mongo(self, value):
         return geojson.GeoJSON.to_instance(value)
 


touilleMan:
That seems good, could you add some tests (especially trying to deserialize dummy/injections in spite of the point field's data, to make sure the geojong lib is strong enough in such case)

Another think: I'd like to make geojson optional

I'm thinking about a mechanism like this:

```python

try:
    import geojson
except ImportError:
    # GeoJSON is not available, Moked PointField

    class NotAvailableField(BaseField, ma_fields.Field):
        REASON = None
        def __init__(self, *args, **kwargs):
            raise NotImplementedError('%s is not available: %s' % type(self).__name__, self.REASON)

    class PointField(NotAvailableField):
        REASON = 'library geojson is not available' 
else:
    # Real PointField definition
    class PointField(BaseField, ma_fields.Field):
       """
       <Point field doc>
       .. note:: Only available if geojson is installed
       """
        ...
```

Of courses this means we need to update the documentation and maybe provide a config in setup.py to be able to do `pip install umongo[geojson]` (however I don't know if we can mix multiple config like `pip install umongo[txmongo][geojson]` so maybe it is to complicated)

What do you think ? Maybe the dep is really small (and there won't be many fields needing dependencies like this one in the future) so we should just keep it simple and ship with the dependency no matter what...

lafrech:
Thanks for the feedback.

I understand you're reluctant to introduce a dependency. It is indeed a really small one, so I'm not sure it is worth the complexity of making it optional. I think it is just big enough to depend on it rather than duplicating what we need from it (which could be pretty much the entire thing). I'll keep it this way for now. I'd rather address the other points first and decide about the dependency when the feature is complete.

Before integration, I'd like to provide the other fields. Or maybe not if I lack time (adding them is easy, adding tests is work), but at least setup everything so that they are easy to add.There should be one abstract generic `GeoField`, from which all other geo fields would inherit (`PointField(GeoField)`, etc.). Should I add this abstract field in abstract.py or in fields.py? I'd say fields.py. Obviously, I wouldn't add it to `__all__` in fields.py. `__all__` would just be added fields that are meant to be used directly. (And now, I'm realizing I forgot to add `PointField`...)

Also, a bit of context about `GeoJSON`, `python-geojson` and validation. When instantiating a `Point` (for instance), `geojson` checks the coordinates are indeed numeric values (`float`, `int`, `Decimal`), and that's pretty much all it does.

It does not check the length of the coordinates array. You may be expecting `[x, y]` and receive `[x, y, z, a, b, c]`.

Someone added an [`is_valid`](https://github.com/frewsxcv/python-geojson/blob/master/geojson/validation.py#L4) method to validate the coordinates a bit further. I can't tell whether or not it covers all incorrect cases as I haven't been looking at geometries other than `Point`. Concerning `Point`, the only thing it checks is the number of coordinates. The test passes if exactly 2 coordinates are provided and fails otherwise.

Since it is a shame to write 

```python
    p = Point(value)
    validate = geojson.is_valid(p)
    if validate.valid == 'no':
        raise ValueError(validate.message)
```

I [proposed](https://github.com/frewsxcv/python-geojson/pull/78) to add a call to `is_valid` in the constructor.

I was soon to realize that the specs do allow more than two parameters. In fact any number of parameters greater than or equal to two is allowed (see my [bug report](https://github.com/frewsxcv/python-geojson/issues/79)). So this validation is flawed: too strict.

Bottom line, calling `is_valid` on a `Point` is both useless and wrong.

Now, assuming we belong to the majority of users who expects two coordinates only, we may want to validate that only two coordinates are provided. This shouldn't be the default case in `geojson` (it is, currently, but it breaks the specs). So we should add a test in our code, or modify the validation on `geojson` side to add a "number of coordinates" parameter. I prefer the latter, but let's see how people there react to my bug report, first.

Maybe a nice feature would be a `_dimension` private attribute that could be set during __init__, would default to `None` but could be set to 2, 3, any number and used at validation to check the number of coordinates.

Oh, and to make things a bit more complicated, MongoDB 2.4 introduced GeoJSON points, but only supporting [x, y] coordinates. [Support for more than two coordinates](https://jira.mongodb.org/browse/SERVER-9220) was introduced in MongoDB 3.0.3. So if we're using MongoDB < 3.0.3, we should not validate `Point`s with more than two coordinates.

lafrech:
Related issue about adding GeoJSON support in PyMongo: https://jira.mongodb.org/browse/PYTHON-1155

lafrech:
If someone wants to pick that up, note that `_serialize` should be modified to work in Mongo world.

```py
 class PointField(BaseField, ma_fields.Field):
 
     def _serialize(self, value, attr, obj):
         if value is None:
             return None
-        return value.coordinates
+        return value['coordinates']
```

Also, with this implementation, mutating a `PointField` does not mark the object as modified. It should inherit from `BaseDataObject`, and it should use setters that call `set_modified`. I'm not sure how much this is compatible with the use of the GeoJSON library. At some point, it might be easier to do everything internally.

Moreover, GeoJSON itself might need better validation. Currently, it is more or less broken. See my comment above for the long story.

If GeoJSON is still considered and adding a dependency is an issue, there could be another repo, like "umongo-extra-fields". (I [just wrote](https://github.com/Scille/umongo/issues/71#issuecomment-263156879) I didn't see the issue with such a dependency, but this lib is a nice example. Currently, its Debian package is slightly outdated. This could be an issue the day someone tries to package umongo.)

Today, I define and use `PointField` in my application code, knowing its limitations, but it can't be integrated as is. I'm going to close this. If I manage to improve it, or if I have anything relevant to add (advice, new limitation discovered,...), I'll be sure to post here about it.

lafrech:
A different approach could be to define `PointField` and his friends as `EmbeddedField`s:

```py
EmbeddedField(PointFieldEmbeddedDoc)
```

where `PointFieldEmbeddedDoc` would be something like

```py
class PointFieldEmbeddedDoc(EmbeddedDocument):

    # TODO: add validator for length
    coordinates = ListField(FloatField)

    # TODO: can only be "Point". Use OneOf validator?
    type = StringField()
```

(Actually, this class would probably inherit from `GeoJsonEmbeddedDoc` so as to share stuff with other `GeoJSON` fields.)

This way we reuse umongo/marshmallow validators and we get mutations caught by umongo.

PR #25, opened 2016-07-13
Add reset_missings parameter to DocumentImplementation's update method

lafrech:
Implementation proposal for the `reset_missings` feature.

Removes value from document if loadable in the Schema but missing in input data.

``` python
user = User(name='Hannibal', team="A-Team")
user.update({'name': 'Looping'})
assert user.team == "A-Team"
user.update({'name': 'Chuck Norris'}, reset_missings=True)
assert user.team == None
```

### DIFF
--- a/umongo/data_proxy.py
+++ b/umongo/data_proxy.py
@@ -80,13 +80,22 @@ class DataProxy:
     def _mark_as_modified(self, key):
         self._modified_data.add(key)
 
-    def update(self, data, schema=None):
+    def update(self, data, schema=None, reset_missings=False):
         schema = schema or self._schema
         # Always use marshmallow partial load to skip required checks
         loaded_data, err = schema.load(data, partial=True)
         if err:
             raise ValidationError(err)
         self._data.update(loaded_data)
+
+        # Set missing values to missing
+        if reset_missings:
+            loadable_fields = set(
+                [k for k, v in schema.fields.items() if not v.dump_only])
+            missing_keys = loadable_fields - set(data)
+            for key in missing_keys:
+                self._data[key] = missing
+
         for key in loaded_data:
             self._mark_as_modified(key)
 
--- a/umongo/data_proxy.py
+++ b/umongo/data_proxy.py
@@ -84,8 +84,7 @@ class BaseDataProxy:
     def update(self, data, reset_missings=False):
 
         # Use marshmallow partial load to skip required checks
-        # unless reset_missings is True
-        loaded_data, err = self.schema.load(data, partial=(not reset_missings))
+        loaded_data, err = self.schema.load(data, partial=True)
         if err:
             raise ValidationError(err)
         self._data.update(loaded_data)
--- a/umongo/data_proxy.py
+++ b/umongo/data_proxy.py
@@ -82,11 +82,6 @@ class DataProxy:
 
     def update(self, data, schema=None, reset_missings=False):
         schema = schema or self._schema
-        # Always use marshmallow partial load to skip required checks
-        loaded_data, err = schema.load(data, partial=True)
-        if err:
-            raise ValidationError(err)
-        self._data.update(loaded_data)
 
         # Set missing values to missing
         if reset_missings:
@@ -97,6 +92,11 @@ class DataProxy:
                 self._data[key] = missing
                 self._mark_as_modified(key)
 
+        # Always use marshmallow partial load to skip required checks
+        loaded_data, err = schema.load(data, partial=True)
+        if err:
+            raise ValidationError(err)
+        self._data.update(loaded_data)
         for key in loaded_data:
             self._mark_as_modified(key)
 
--- a/umongo/data_proxy.py
+++ b/umongo/data_proxy.py
@@ -88,22 +88,23 @@ class DataProxy:
     def update(self, data, schema=None, reset_missings=False):
         schema = schema or self._schema
 
-        # Set missing values to missing
-        if reset_missings:
-            loadable_fields = set(
-                [k for k, v in schema.fields.items() if not v.dump_only])
-            missing_keys = loadable_fields - set(data)
-            for key in missing_keys:
-                self._data[key] = missing
-                self._mark_as_modified(key)
-
-        # Always use marshmallow partial load to skip required checks
-        loaded_data, err = schema.load(data, partial=True)
+        # Use marshmallow partial load to skip required checks
+        # unless reset_missings is True
+        loaded_data, err = schema.load(data, partial=(not reset_missings))
         if err:
             raise ValidationError(err)
         self._data.update(loaded_data)
         for key in loaded_data:
             self._mark_as_modified(key)
+
+        # Delete missing fields unless dump_only or with a default value
+        if reset_missings:
+            deletable_fields = set(
+                [k for k, v in schema.fields.items()
+                 if not v.dump_only and v.missing is missing])
+            missing_keys = deletable_fields - set(data)
+            for key in missing_keys:
+                self.delete(key)
 
     def load(self, data, partial=False, schema=None):
         schema = schema or self._schema
--- a/umongo/data_proxy.py
+++ b/umongo/data_proxy.py
@@ -98,11 +98,12 @@ class BaseDataProxy:
         # Delete missing fields unless dump_only or with a default value
         if reset_missings:
             deletable_fields = set(
-                [k for k, v in self.schema.fields.items()
+                [k for k, v in self._fields.items()
                  if not v.dump_only and v.missing is missing])
             missing_keys = deletable_fields - set(data)
             for key in missing_keys:
                 self.delete(key)
+                self.not_loaded_fields.discard(self._fields[key])
 
     def load(self, data, partial=False):
         # Always use marshmallow partial load to skip required checks
--- a/umongo/data_proxy.py
+++ b/umongo/data_proxy.py
@@ -89,7 +89,7 @@ class DataProxy:
         schema = schema or self._schema
 
         # Use marshmallow partial load to skip required checks
-        # unless reset_missings is True
+        # unless reset_missings is True
         loaded_data, err = schema.load(data, partial=(not reset_missings))
         if err:
             raise ValidationError(err)
--- a/umongo/data_proxy.py
+++ b/umongo/data_proxy.py
@@ -95,6 +95,7 @@ class DataProxy:
             missing_keys = loadable_fields - set(data)
             for key in missing_keys:
                 self._data[key] = missing
+                self._mark_as_modified(key)
 
         for key in loaded_data:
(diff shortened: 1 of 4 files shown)


touilleMan:
Can you add some unittests to this ?

umongo/data_proxy.py
```diff
@@ -80,13 +80,23 @@ def dump(self, schema=None):
...
+        # Set missing values to missing
+        if reset_missings:
+            loadable_fields = set(
+                [k for k, v in schema.fields.items() if not v.dump_only])
+            missing_keys = loadable_fields - set(data)
+            for key in missing_keys:
+                self._data[key] = missing
+                self._mark_as_modified(key)
+
         for key in loaded_data:
```
touilleMan: This for loop should stay after `self._data.update(loaded_data)` given it has nothing to do with the reset_missings logic

lafrech: Right. And in fact, the whole "# Set missing values to missing" block should be above the "loaded_data, err = schema.load(data, partial=True)" so that a missing value with a default value is set to `missing`, not removed.

lafrech:
I reworked the feature and added some tests (I don't understand why coveralls reports a coverage decrease).

When using `reset_missings`

- If a required field is missing, an exception is raised.
- If a field has a default value, it is set to default.

I wasn't sure how to deal with the "default value" case. An alternative way would be to proceed in three steps, but it seemed less readable:

```py
        # 1 - Always do that first to raise en Exception on wrong data before modifying anything
        # Use marshmallow partial load to skip required checks
        # unless reset_missings is True
        loaded_data, err = schema.load(data, partial=(not reset_missings))
        if err:
            raise ValidationError(err)

        # 2 - Don't mind default values
        # Delete missing fields unless dump_only
        if reset_missings:
            deletable_fields = set(
                [k for k, v in schema.fields.items()
                 if not v.dump_only])
            missing_keys = deletable_fields - set(data)
            for key in missing_keys:
                self.delete(key)

        # 3 - Do that afterwards to set default values (this overrides delete from above step)
        self._data.update(loaded_data)
        for key in loaded_data:
            self._mark_as_modified(key)
```

Thank you for your feedback.

umongo/data_proxy.py
```diff
@@ -85,16 +85,27 @@ def dump(self, schema=None):
...
             raise ValidationError(err)
         self._data.update(loaded_data)
         for key in loaded_data:
             self._mark_as_modified(key)
 
+        # Delete missing fields unless dump_only or with a default value
+        if reset_missings:
+            deletable_fields = set(
+                [k for k, v in schema.fields.items()
+                 if not v.dump_only and v.missing is missing])
```
touilleMan: Not sure about the `v.missing is missing` filter...
I would say all field should be reset, either back to `missing` object or to their custom `field.missing` value (or to `field.missing()` if it's a callable)

touilleMan:
> I don't understand why coveralls reports a coverage decrease

umongo is pretty special because you have 3 diferents frameworks that need to be test against.
Thus coveralls get the code coverage in multiple parts (i.g. first the tests for pymongo ends up and coveralls get ~70% coverage, then the tests for motor are done and coveralls get another 70%, and finally txmongo is done. In the end you get 94%, but at one point coveralls show the coverage was only 70%...)

touilleMan:
> I wasn't sure how to deal with the "default value" case. An alternative way would be to proceed in three steps, but it seemed less readable:

What about this ?
```python
        loaded_data, err = schema.load(data, partial=(not reset_missings))
        if err:
            raise ValidationError(err)

        if reset_missings:
            deletable_fields = set(
                [k for k, v in schema.fields.items() if not v.dump_only])
            for key in deletable_fields - set(data):
                field = schema.fields[key]
                self.set(key, field.missing)

        self._data.update(loaded_data)
        for key in loaded_data:
            self._mark_as_modified(key)
```

lafrech:
> What about this ?

Apparently, it fails when setting an integer field to `missing`.

```
tests/test_data_proxy.py:340: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
umongo/data_proxy.py:103: in update
    self.set(key, field.missing)
umongo/data_proxy.py:156: in set
    value = field._deserialize(value, name, None)
../.venvs/umongo/lib/python3.4/site-packages/marshmallow-2.9.1-py3.4.egg/marshmallow/fields.py:634: in _deserialize
    return self._validated(value)
../.venvs/umongo/lib/python3.4/site-packages/marshmallow-2.9.1-py3.4.egg/marshmallow/fields.py:619: in _validated
    self.fail('invalid')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <fields.IntegerField(default=<marshmallow.missing>, attribute='e', validate=None, required=False, load_only=Fal...s} must be unique together.', 'unique': 'Field value must be unique.'}, io_validate=None, unique=False, instance=No
key = 'invalid', kwargs = {}, msg = 'Not a valid integer.'

    def fail(self, key, **kwargs):
        """A helper method that simply raises a `ValidationError`.
            """
        try:
            msg = self.error_messages[key]
        except KeyError:
            class_name = self.__class__.__name__
            msg = MISSING_ERROR_MESSAGE.format(class_name=class_name, key=key)
            raise AssertionError(msg)
        if isinstance(msg, basestring):
            msg = msg.format(**kwargs)
>       raise ValidationError(msg)
E       marshmallow.exceptions.ValidationError: Not a valid integer.

../.venvs/umongo/lib/python3.4/site-packages/marshmallow-2.9.1-py3.4.egg/marshmallow/fields.py:218: ValidationError
```

> Not sure about the v.missing is missing filter...

Yeah, I don't like it either. Hence my proposal to put step 3 after step 2 (see my comment). But none of these options looked so obvious and simple it had to be the right one.

> I would say all field should be reset, either back to missing object or to their custom field.missing value (or to field.missing() if it's a callable)

Yes, I also considered this option, but then we'd have to reproduce the ["if callable" test](https://github.com/Scille/umongo/blob/master/umongo/data_proxy.py#L193) logic you'd like to get rid of ("# TODO: we should be able to do that by configuring marshmallow...").

I thought this was going to be a simple change, and now I feel awkward about it. I have the feeling I'm walking on Marshmallow territory, messing up with `missing`. You're probably right it deserves more thinking (https://github.com/Scille/umongo/issues/36).

lafrech:
I modified the behavior for an update with `reset_missings` and a missing required field.

Returning a validation error is problematic.

We have a use case with a required password field that is updated using a method (for the hash/salt process) and is not in the data kwargs passed to update. The update takes place in two times: first update all fields, then update password, then commit. I don't want an exception if the required password field is not in the update data.

This use case may be a bad pattern, but anyway, the change is coherent with the fact that required constraint is validated at commit time, so one can instantiate an object without all the required data at `__init__` time.

Although I think this feature is needed, I'm not comfortable with this implementation. I hope https://github.com/Scille/umongo/issues/36 will help clarifying this.

lafrech:
I've been thinking this through and I ended up taking another path.

It all came from the need to update an object easily when dealing with a `PUT` request in our API.

When things are simple, you have a Document that maps straight to an API resource, you get the API schema using `as_marshmallow_schema` and you're all set. In this case, in a `PUT`, you don't want to just update the fields in the request payload (that would be some sort of `PATCH`), you want to reset the fields that are missing. Otherwise, you just can't use `PUT` to remove a field.

If you want to avoid the hassle of writing and maintaining such code

```py
    document.attr1 = request_doc.attr1
    document.attr2 = request_doc.attr2
    ...
```

you need a function/method that knows which fields are expected in the `PUT` request so that it can automatically remove the missing ones.

That's when that `reset_missings` feature kicked in.

Neat, but it's got at least one serious shortcoming.

Again, when things are simple, the API schema is so close to the umongo schema the list of fields is the same. However, when you customize the API schema generated with `as_marshmallow_schema` to remove some fields, the model does not know about that. So when using `reset_missings` in `update`, you end up removing the fields that you excluded from the API, because the `Document` doesn't know they're not expected if they are not `read_only`.

Basically, it works only if the fields are marked as `read_only` in the model, not if they are stripped of the API schema afterwards.

I'm not sure exactly what `read_only` is supposed to mean on model side, apart from being passed down to the marshmallow schema. Apparently, when setting a field `read_only`, you can't use it in load (that includes `Document` instantiation such as `Document(**{'ro_field': 'val'})`) but you can set it using `setattr`: `document.ro_field = val`. This looks inconsistent, but I haven't taken the time to investigate it, and it is another story.

In the general case, things are not simple. The API is not the model. The resource is not the document. Therefore, deciding which fields should be considered missing is up to the API, not the model. So I ended up coding a function that updates the document removing fields missing from the API schema, not the model schema.

The way I did it, it is a module function that takes the `Document`, the values and the `Schema` as argument. It could be bound to the `Document` or maybe the `Schema`. I haven't taken the time to do that. It can be done without modifying umongo (using an abstract `Document`, or using `base_schema_cls` of `as_marshmallow_schema`).

I'll post the code here in case anybody is interested. Nothing to be thrilled about.

**TL; DR**

I'm not using this feature right now and since it's bugging me almost every time I rebase on latest master, I'll be dropping it from my umongo fork.

I don't think I'll work on it any more.

Maybe someone will want to pick it up, as it may have an interest other than the use case I wanted it for.

Thank you for looking into it and providing valuable feedback.

lafrech:
> I'll post the code here in case anybody is interested.

Here it is:

```py
def update_document_with_schema(document, values, schema):
    """Update a umongo Document using provided Schema

       While updating, for all fields in the Schema that are not read-only,
       the corresponding attribute is removed if the value is not in values.
    """

    # Ensure we have a Schema instance, not a class
    # so that we ignore excluded fields
    if isinstance(schema, Schema.__class__):
        schema = schema()

    # Raise an error if unexpected values are passed
    loadable_fields = {k: v for k, v in schema.fields.items()
                       if not v.dump_only}
    unknown_fields = {k for k, v in values.items()
                      if k not in loadable_fields.keys()}
    if unknown_fields:
        raise ValidationError(
            ['Unknown field name {field}.'.format(field=field)
             for field in unknown_fields])

    # Update document, deleting missing values
    for name, field in loadable_fields.items():
        if name in values:
            val = values.get(name)
            setattr(document, name, val)
        else:
            delattr(document, name)
```

touilleMan:
```python
    if isinstance(schema, Schema.__class__):
```
You'd better use `issubclass(schema, Schema)` instead ^^

PR #26, opened 2016-07-13
Fix $unset syntax in `_to_mongo_update`

lafrech:
I'm having issues unsetting fields. It it likely I misunderstood the whole thing, but from what I understand, `_to_mongo_update` generates `$unset` operations as (see [here](https://github.com/Scille/umongo/blob/master/tests/test_data_proxy.py#L131) for instance):

    {'$unset': ['a', 'b']}

This generates errors:

    "Modifiers operate on fields but we found a Array instead. For example: {$mod: {<field>: ...}} not {$unset: [ \"a\", \"b\" ]}"

Indeed, the [docs](https://docs.mongodb.com/manual/reference/operator/update/unset/) indicate the expected syntax is:

    $unset: { quantity: "", instock: "" }

I'm a bit surprised no one ever hit this before, which makes me think I might be wrong.

Anyway, here's a fix proposal to illustrate the idea. Obviously, it breaks the tests. I'd rather you validate the idea before modifying the tests.

Also, I don't understand the use of `sorted` in the original code.

### DIFF
--- a/tests/test_data_proxy.py
+++ b/tests/test_data_proxy.py
@@ -69,7 +69,7 @@ class TestDataProxy(BaseTest):
         assert d.to_mongo(update=True) is None
         d.set('a', 3)
         d.delete('b')
-        assert d.to_mongo(update=True) == {'$set': {'a': 3}, '$unset': ['in_mongo_b']}
+        assert d.to_mongo(update=True) == {'$set': {'a': 3}, '$unset': {'in_mongo_b': ''}}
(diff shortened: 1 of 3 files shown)


touilleMan:
You're right about this, please fix the tests and I'll merge this PR

The `sorted` was here to make testing easier by always returning the same list

lafrech:
Done!

touilleMan:
:+1:

PR #41, opened 2016-08-25
Add is_modified method to Document

lafrech:
I would like to be able to tell if a Document was modified.

Was there a simpler / more recommended way?

I suppose we could also use 

    if self._data._to_mongo_update() is not None

but it takes more computations.

I could add tests if you're interested.

### DIFF
--- a/tests/test_document.py
+++ b/tests/test_document.py
@@ -2,7 +2,7 @@ import pytest
 from datetime import datetime
 from bson import ObjectId, DBRef
 
-from umongo import Document, Schema, fields, exceptions
+from umongo import Document, EmbeddedDocument, Schema, fields, exceptions
 
 from .common import BaseTest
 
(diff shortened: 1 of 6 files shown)


touilleMan:
This is a pretty good idea ! Beside it adds coherence with `EmbeddedDocument`.

Can you provide tests for this in `test_document.py` ?

lafrech:
Great. I'll try to do that.

lafrech:
I just added a few tests.

tests/test_document.py
```diff
@@ -230,6 +230,36 @@ class CustomIdInheritance(CustomId):
...
+        @self.instance.register
+        class Vehicle(EmbeddedDocument):
+            name = fields.StrField()
+
+        @self.instance.register
+        class Driver(Document):
+            name = fields.StrField()
+            vehicle = fields.EmbeddedField(Vehicle)
+
+        driver = Driver(name='Marty')
```
touilleMan: This document has not been commited yet, thus it should be `is_modified() == True`

You should check `is_created` within `is_modified` for this.

touilleMan:
Added a comment about the need to use `is_created` inside `is_modified`

lafrech:
> Added a comment about the need to use is_created inside is_modified

Yep. I get notified of those comments as well.

> This document has not been commited yet, thus it should be is_modified() == True
> 
> You should check is_created within is_modified for this.

Right. I pushed a fix.

touilleMan:
Good job ! :+1:

Issue #42, opened 2016-08-29
ReferenceField lazy/auto dereference


lafrech:
Currently, a `ReferenceField` deserializes as (for instance) a `PyMongoReference`, and must be dereferenced manually using `fetch()`.

```py
from pymongo import MongoClient
from umongo import Instance, Document, fields

db = MongoClient().test
instance = Instance(db)

@instance.register
class Ref(Document):
    a = fields.IntField()

@instance.register
class Doc(Document):
    ref = fields.ReferenceField(Ref)
    
ref = Ref(a=12)
ref.commit()

# Here's my object
print(ref)
# <object Document __main__.Ref({'a': 12, '_id': ObjectId('57c4ae2cf66737156a9ebfff')})>

# Let's reference it in another object
doc = Doc(ref=ref)

# I can't get it directly from here
print(doc.ref)
# <object umongo.frameworks.pymongo.PyMongoReference(document=Ref, pk=ObjectId('57c4ae2cf66737156a9ebfff'))>

# I need to call fetch manually
print(doc.ref.fetch())
# <object Document __main__.Ref({'a': 12, '_id': ObjectId('57c4ae2cf66737156a9ebfff')})>
```

This is not as practical as automatic dereferencing (systematic or lazy as in MongoEngine).

Is this a design choice (simplicity) or something that you'd like to see improved?

In practice, I could call `fetch()` every time I'm accessing a ref field, but there are cases where you need to pass an object to another method you can't modify (imported library) and you really want `doc.ref` to be the referenced document, not a `PyMongoReference`.

```py
print(doc.ref.a)
AttributeError: 'PyMongoReference' object has no attribute 'a'
```

And I can't even dereference every reference manually to get a "clean" object:

```py
# This won't work
doc.ref = doc.ref.fetch()
print(doc.ref)
# <object umongo.frameworks.pymongo.PyMongoReference(document=Ref, pk=ObjectId('57c4ae2cf66737156a9ebfff'))>
```

I'm kinda stuck. Is there a way around this? Am I missing something?

touilleMan:
> Is this a design choice (simplicity) or something that you'd like to see improved?

It's a design choice !
Given umongo support asynchronous drivers, retrieving data from the database must be explicit in order to yield on this operation:

```python
# using asyncio
async def get_ref_a(doc):
    referenced = await doc.ref.fetch()
    return referenced.a
```

[...]

PR #43, opened 2016-08-30
ReferenceField lazy dereference

lafrech:
Here's an attempt to add dereferencing feature to `Reference` (see https://github.com/Scille/umongo/issues/42).

It's pretty hackish... Basically, the idea is that every attribute or item access to `Reference` is redirected to the referenced document, except for attributes in `Reference`'s `__dict__`.

I tried to figure out a better way adding a `__get__` method to `Reference` so that it returns `self.fetch()` when called from another class, but I didn't get it to work.

### DIFF
--- a/umongo/data_objects.py
+++ b/umongo/data_objects.py
@@ -104,13 +104,9 @@ class Reference:
     error_messages = I18nErrorDict(not_found='Reference not found for document {document}.')
 
     def __init__(self, document_cls, pk):
-
-        # Attribute getter/setter/deleter are overridden to provide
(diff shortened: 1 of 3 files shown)


lafrech:
There is an error in `__setattr__`:

```py
    def __setattr__(self, name, value):
        if name in self.__dict__:
            self.__dict__[name] = value
        else:
-            setattr(self.fetch(), name)
+            setattr(self.fetch(), name, value)
```

I don't think I'll be working on this, so I'm closing this for now.

PR #45, opened 2016-09-01
DataProxy: Empty not_loaded_fields on non-partial loads

lafrech:
I may be mistaken, but the way I understand the code, `self.not_loaded_fields` is not reset when loading non-partial, so if a partial load was performed previously, it still has fields missing from this previous load.

### DIFF
--- a/tests/test_data_proxy.py
+++ b/tests/test_data_proxy.py
@@ -303,3 +303,16 @@ class TestDataProxy(BaseTest):
         assert d.get('normal') is missing
         assert d.get('loaded') == "foo"
         assert d.get('loaded_but_empty') == missing
+
+        # Partial, then not partial
+        d = DataProxy(MySchema())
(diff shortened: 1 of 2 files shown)


lafrech:
Likewise, maybe we should do something in `DataProxy.update()` to update `not_loaded_fields`. I new fields have been loaded, they should be removed from `not_loaded_fields`, right?

touilleMan:
merged

lafrech:
Great.

Should we also fix `update`?

Something like this (untested):

```py
    def update(self, data):
        # Always use marshmallow partial load to skip required checks
        loaded_data, err = self.schema.load(data, partial=True)
        if err:
            raise ValidationError(err)
        self._data.update(loaded_data)
+       self.not_loaded_fields = tuple(
+            set(self.not_loaded_fields) - set(self._fields[k] for k in data))
        for key in loaded_data:
            self._mark_as_modified(key)
```

touilleMan:
you're right, commit 1dd4076 take care of this now 👍

PR #47, opened 2016-09-05
Fix PyMongo WrappedCursor slicing

lafrech:
This allows slicing umongo's PyMongo WrappedCursor like you would slice a PyMongo Cursor.

```py
    cursor = Student.find()
    for elem in cursor[2:5]:
        ...
```


touilleMan:
nice !

Issue #35, opened 2016-07-27
Add `conditions` to delete method


touilleMan:
`commit` method has a `conditions` argument (https://github.com/Scille/umongo/blob/master/umongo/frameworks/pymongo.py#L66), should be also available for `delete` method

touilleMan:
merged

PR #49, opened 2016-09-09
 Add `conditions` to delete method

lafrech:
Trying to address https://github.com/Scille/umongo/issues/35.

Incidentally, in motor_asyncio, `pre_delete` is now executed only if doc is created, like on the other frameworks.

Also added possibility for `pre_delete` to add additional query filters, like `pre_update`.


lafrech:
I think the error is due to Travis.

I can squash the commits if that matters to you.

touilleMan:
Merged ;-)

PR #51, opened 2016-09-19
Fix update not loaded fields

lafrech:
See https://github.com/Scille/umongo/pull/45#issuecomment-244017493:

> Likewise, maybe we should do something in DataProxy.update() to update not_loaded_fields. If new fields have been loaded, they should be removed from not_loaded_fields, right?


lafrech:
Done in https://github.com/Scille/umongo/commit/1dd4076998c4171e2482bd4609933c761311e2e2.

PR #52, opened 2016-09-19
Fix dataproxy update not_loaded_fields

lafrech:
See comment: https://github.com/Scille/umongo/commit/1dd4076998c4171e2482bd4609933c761311e2e2#commitcomment-19075700

> I think it should be
> 
>     for k in data:
> 
> otherwise it chokes on fields with attribute.


touilleMan:
Sorry I didn't see this PR and fixed this with 6b01362

PR #59, opened 2016-10-13
Serialization issue with DictField

lafrech:
I have an issue with the introduction of `UserDict`. See this PR for test case and workaround.

Basically, when I export a schema with a `DictField` as a Marshmallow `Schema`, the `Dict` field serializes as a `UserField`, because `_serialize` in `Dict` is noop. Unfortunately, `UserField` does not really behave like a dict. `isinstance(UserDict(), dict)` returns `False`. Besides, `json.dumps()` will throw a `TypeError` on a `UserDict`.

```python
    import json
    from collections import UserDict
    json.dumps(UserDict(a=42))
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/usr/lib/python3.4/json/__init__.py", line 230, in dumps
        return _default_encoder.encode(obj)
      File "/usr/lib/python3.4/json/encoder.py", line 192, in encode
        chunks = self.iterencode(o, _one_shot=True)
      File "/usr/lib/python3.4/json/encoder.py", line 250, in iterencode
        return _iterencode(o, 0)
      File "/usr/lib/python3.4/json/encoder.py", line 173, in default
        raise TypeError(repr(o) + " is not JSON serializable")
    TypeError: {'a': 42} is not JSON serializable
```

I had a hard time figuring this out, and I'm not sure where it should be addressed.

A real `dict` can be accessed:

```py
    json.dumps(UserDict(a=42).data)
    '{"a": 42}'
    json.dumps(dict(UserDict(a=42)))
    '{"a": 42}'
```

and this is how I worked this around (see PR).

Feedback welcome.

Enjoy PyCon!

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -1,3 +1,4 @@
+import json
 from bson import ObjectId, DBRef
 import pytest
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


touilleMan:
Digging into this I realised there was no reason any more for Dict and List not to inherit from the builtins types (the reasons used to be Dict and List use metaclasses which cannot be mixed with base types)

So b502900 should put an end to your troubles ^^

PR #60, opened 2016-10-14
Make inheritance _cls field load_only

lafrech:
Shouldn't _cls field be load-only?

AFAIU, this is internal stuff. It does not have to be exposed through the API when serializing the data.

### DIFF
--- a/umongo/schema.py
+++ b/umongo/schema.py
@@ -30,7 +30,7 @@ def on_need_add_id_field(bases, fields):
 
 def add_child_field(name, fields):
     from .fields import StrField
-    fields['cls'] = StrField(attribute='_cls', missing=name, dump_only=True)
+    fields['cls'] = StrField(attribute='_cls', missing=name, dump_only=True, load_only=True)
 
 
 class Schema(BaseSchema):


touilleMan:
The thing is, given `_cls` is used to determine the type of the document, it can be nice to provide this information through your API.
The same way, you can have an API that allow to create different types of document, which means the user should be able to provide the `_cls` field.

lafrech:
Indeed, we want to provide this information via the API for the two symmetric uses cases you're presenting, but the name of the class is internal stuff. We're exposing the type with a nicer string.

(To do so, we had to use a Marshmallow extension that provides this feature: https://github.com/maximkulkin/marshmallow-oneofschema. See discussion here: https://github.com/marshmallow-code/marshmallow/issues/412)

touilleMan:
I didn't know about marshsmallow-onofschema, pretty interesting ;-)

I've just realized your usecase could be really easily corrected by allowing the use of marshmallow's pre/post processors, so I've added this feature (see 14f6cea)

example from the userguide:
```python
    >>> @instance.register
    ... class Dog(Document):
    ...     name = fields.StrField(required=True)
    ...     breed = fields.StrField(missing="Mongrel")
    ...     birthday = fields.DateTimeField()
    ...     @marshmallow.post_dump
    ...     def customize_dump(self, data):
    ...         data['name'] = data['name'].capitalize()
    ...         data['brief'] = "Hi ! My name is %s and I'm a %s" % (data['name'], data['breed'])"
    ...
    >>> Dog(name='scruffy').dump()
    {'name': 'Scruffy', 'breed': 'Mongrel', 'brief': "Hi ! My name is Scruffy and I'm a Mongrel"}
```

lafrech:
I'm not sure this covers our use case.

- [x] When serializing, we want to write the type of the data in the API. This is a string. It could be the class name, although we want something else and your example shows a way to do that.

- [ ] When deserializing, we want to get the type from the input data and select the right deserialization schema accordingly. I'm not sure we can do this with a pre-load action, can we?

My comment was a bit simplistic. We didn't introduce marshmallow-oneofschema just to rename the type attribute, but because it provides the logic both ways ([|de]serialization), including selection of the right schema when deserializing.

(It's only 75 loc, anyway, so not a huge dependency.)

touilleMan:
> When deserializing, we want to get the type from the input data and select the right deserialization schema accordingly. I'm not sure we can do this with a pre-load action, can we?

For this kind of usecase, I would go KISS by creating a factory which would select the right Document class to use for deserialization:
```python
@instance.register
class Animal(Document):
    ...

@instance.register
class Dog(Animal):
    ...

def load_from_type(data, root_cls=None):
    if 'type' not in data:
        # raise error here
    type_name = data.pop('type')
    if root_cls and type_name not in root_cls.opts.children:
        # raise error here
    try:
        doc_cls = instance.retrieve_document(type_name)
    except NotRegisteredDocumentError:
        # raise error here
    return doc_cls(**data)
```

usage:
```python
>>> load_from_type({'type': 'Dog', 'name': 'Rox'})
<Dog({'name': 'Rox'})>
>>> load_from_type({'type': 'Stone', ...})
Execption "Stone is not a children of Animal"
```

lafrech:
I'm rethinking this. I understand the use for `cls` to be in the `dump`, especially if it is expected at `load` time to select the right child class.

If I want my API to hide this, I should act upon my API schemas to exclude this field.

I'll reopen if needed, but for now, please ignore.

Thank you for your support.

PR #63, opened 2016-10-25
DataProxy.set(): don't call validation if allow_none = True and value is None

lafrech:
Current implementation of `DataProxy.set()` does not allow setting `None` in a field with `allow_none = True` and validators defined, because it calls validation on the value even if it is `None`.

In Marshmallow, the validators are not called in this case (see https://github.com/marshmallow-code/marshmallow/blob/e6ddef938b3b274a4965cad344c3864e643e6baa/marshmallow/fields.py#L254).

My explanation sucks, but hopefully, the tests will speak for themselves...

### DIFF
--- a/umongo/data_proxy.py
+++ b/umongo/data_proxy.py
@@ -145,7 +145,7 @@ class BaseDataProxy:
     def set(self, name, value, to_raise=KeyError):
         name, field = self._get_field(name, to_raise)
         value = field._deserialize(value, name, None)
-        if getattr(field, 'allow_none', False) is False or value is not None:
+        if value is not None or not getattr(field, 'allow_none', False):
             field._validate(value)
         self._data[name] = value
         self._mark_as_modified(name)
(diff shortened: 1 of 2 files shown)


umongo/data_proxy.py
```diff
@@ -145,7 +145,8 @@ def get(self, name, to_raise=KeyError):
     def set(self, name, value, to_raise=KeyError):
         name, field = self._get_field(name, to_raise)
         value = field._deserialize(value, name, None)
-        field._validate(value)
+        if getattr(field, 'allow_none', False) is False or value is not None:
```
touilleMan: small optim: I would switch the checks given None compare is really cheaper than `getattr`:
` if value is not None or getattr(field, 'allow_none', False) is False:`

lafrech: Good point. I updated the PR.

I also changed 

    getattr(field, 'allow_none', False) is False

into

    not getattr(field, 'allow_none', False)

I can revert if you prefer the explicit `is False` form.

touilleMan:
:+1:

PR #67, opened 2016-11-02
Failing tests for EmbeddedDocument issues

lafrech:
Hopefully, this will help you get an idea of the issues I'm facing.

I purposely left the debug `print` statements.

You'll see that in `EmbeddedField._deserialize`, the deserialization is done recursively and in the end `_deserialize_from_mongo` is called with an object that is not a `dict` but an `Implementation` class.

Then in `EmbeddedDocument.build_from_mongo`, the `'_cls in data'` test raises an exception because `key in some_embedded_doc_instance` is invalid.

I don't claim I have the whole picture clear, but I still think the root cause is calling "from_mongo" stuff to do OO deserialization.

I hope the testcase helps, but frankly, I'd start by wondering about the "from_mongo" stuff in `_deserialize` as the whole thing may unravel from there.

My hacky-fix branch in https://github.com/Scille/umongo/pull/56 passes the tests. The implementation may have issues as well, so it is not meant to be integrated as is, but it may shed light on my explanation and point where the issue can be.

Please ask for clarification is this is still too blurry.

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -180,9 +180,14 @@ class TestFields(BaseTest):
     def test_embedded_document(self):
 
         @self.instance.register
+        class OtherEmbedded(EmbeddedDocument):
+            f = fields.IntField()
+
+        @self.instance.register
         class MyEmbeddedDocument(EmbeddedDocument):
             a = fields.IntField(attribute='in_mongo_a')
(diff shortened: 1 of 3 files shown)


lafrech:
Fixed in https://github.com/Scille/umongo/pull/84.

PR #69, opened 2016-11-10
Fix allow_none fields [de|]serialization

lafrech:
I'm afraid there's an issue with `allow_none` fields.

For both serialization and deserialization to/from Mongo, if the value is None, None should be returned and all `_(de)serialize_from_mongo` overridden methods should be skipped, as they don't deal with the `allow_none` case and either break (e.g. StrictDateTime) or return a nullish object (empty list, empty dict).

See attached fix proposal.

I didn't code any test as I'm in a bit of a rush and I'd rather have your feedback first.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,7 @@ History
 
 dev
 ---
+* Fix fields serialization/deserialization when allow_none is True (see #69)
 * Fix ReferenceFild assignment from another ReferenceField (see #110)
 * Fix deletion of field proxied by a property (see #109)
 * Fix StrictDateTime bonus field: _deserialize does not accept datetime.datetime
--- a/tests/test_embedded_document.py
+++ b/tests/test_embedded_document.py
@@ -28,7 +28,6 @@ class TestEmbeddedDocument(BaseTest):
 
         assert document is not None
 
-
     def test_embedded_document(self):
         @self.instance.register
         class MyEmbeddedDocument(EmbeddedDocument):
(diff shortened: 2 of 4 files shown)


lafrech:
Do you agree with the issue and the fix proposal? Should I add tests for this?

lafrech:
I just added a few test cases to demonstrate the issue.

Most fields manage the `allow_none` case correctly because they don't override `_serialize_to_mongo` and `_deserialize_from_mongo` from `BaseField`, which pass `None` transparently.

However, special fields that do override those methods (`DictField`, `ListField`, `ReferenceField`, `GenericReferenceField` and `EmbeddedField`) don't manage `allow_none` as expected (see the tests) and either remove the field or even crash.

Since `allow_none` is allowed for those fields, it should be managed correctly.

I see to approaches to this. Either manage the `None` case in each individual field, either factorize this logic in `serialize_to_mongo` and `deserialize_from_mongo` in `BaseField` (what I did in this PR).

I'm not sure which approach is best, but I don't think it hurts to integrate this as is. Feedback welcome.

Happy new year, by the way.

lafrech:
I just rebased for cleaner history. Any objection to this?

lafrech:
Rebased.

lafrech:
Anything wrong with this? Anything you'd like me to rework?

We're using this patch in our application.

touilleMan:
Sorry, I totally missed this PR :'-(

I'll try to have a look on this this week...

lafrech:
Rebased + history updated.

lafrech:
Rebased again.

Anything you dislike in this change? In the tests?

touilleMan:
Sorry for taking so long, this PR was totally fine ^^

PR #70, opened 2016-11-15
Don't fail if missing embedded doc with required fields

lafrech:
Please tell me what you think about this.

Currently, required fields validation fails if an embedded document with required fields is missing, even if the embedded document itself is not required.

I think this should fail only if the embedded document itself is required.

Consider this model:

```
User
  - name (string, required)

Team
  - manager (embedded user, not required)
```

manager field in Team is not required. If a manager is provided, it should have a name (required field), but if no manager is provided, I don't think validation should fail. No manager, no problem!

It should fail only 

- if manager is required 

- or if manager is provided without name, whether or not it is required.


touilleMan:
I totally agree on this behavior :+1: 

I Changed a bit your PR before merge it to clean tests (new `validate_required` allows generic test on required field instead of having one test per framework)

PR #72, opened 2016-11-22
Fix required check on update

lafrech:
Looks like there's a `self.required_validate()` missing in `commit` (all drivers) when updating a document.

Fix and tests in the PR.

The first commit added tests about embedded field and list of embedded fields which is unnecessary to test this particular issue, so I "simplified them out". I left them in a separate commit so that they're easy to grab if you find them useful anyway.


touilleMan:
Good catch !

PR #79, opened 2016-12-06
EmbeddedField('string') fails if corresponding EmbeddedDocument is not registered yet

lafrech:
I'm defining an EmbeddedDocument like this:

```py
@db_instance.register
class User(EmbeddedDocument):
    score = fields.EmbeddedField('Score')
```

I provide `Score` as as string because 

- either at import time, the module defining it is not imported yet (and I'd rather avoid it for circular import reasons)

- either it is defined further in the file

Unfortunately, it fails because umongo tries to resolve the EmbeddedDocument right away (no lazyness).

I'm pretty sure you introduced the possibility to use a string precisely for those use cases, right? So I'd call it a bug.

This PR provides a simple failing test.

At import time, `_patch_field` stuff from the builder tries to access `nested` property, which in turns calls `embedded_document_cls` property, which calls `retrieve_embedded_document`.

Traceback:

```
___________________________________________________________________________________________ TestFields.test_embedded_field ___________________________________________________________________________________________

self = <tests.test_fields.TestFields object at 0x7f4cb00f2e10>

    def test_embedded_field(self):
    
        @self.instance.register
        class EmbeddedDocument_A(EmbeddedDocument):
            field = fields.IntField()
    
        @self.instance.register
>       class EmbeddedDocument_B(EmbeddedDocument):

tests/test_fields.py:329: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
umongo/instance.py:103: in register
    implementation = self._register_embedded_doc(template)
umongo/instance.py:117: in _register_embedded_doc
    implementation = self.builder.build_embedded_document_from_template(template)
umongo/builder.py:279: in build_embedded_document_from_template
    schema_cls = self._build_schema(template, schema_bases, schema_nmspc)
umongo/builder.py:200: in _build_schema
    self._patch_field(field)
umongo/builder.py:194: in _patch_field
    for embedded_field in field.schema.fields.values():
../../../.virtualenvs/proleps/lib/python3.4/site-packages/marshmallow/fields.py:391: in schema
    if isinstance(self.nested, SchemaABC):
umongo/fields.py:390: in nested
    return self.embedded_document_cls.Schema
umongo/fields.py:404: in embedded_document_cls
    self.embedded_document)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <umongo.instance.Instance object at 0x7f4cb00f2cf8>, name_or_template = 'EmbeddedDocument_C'

    def retrieve_embedded_document(self, name_or_template):
        """
            Retrieve a :class:`umongo.embedded_document.EmbeddedDocumentImplementation`
            registered into this instance from it name or it template class
            (i.e. :class:`umongo.EmbeddedDocument`).
            """
        if not isinstance(name_or_template, str):
            name_or_template = name_or_template.__name__
        if name_or_template not in self._embedded_lookup:
            raise NotRegisteredDocumentError(
>               'Unknown embedded document class `%s`' % name_or_template)
E           umongo.exceptions.NotRegisteredDocumentError: Unknown embedded document class `EmbeddedDocument_C`

umongo/instance.py:67: NotRegisteredDocumentError
```

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -8,6 +8,7 @@ from uuid import UUID
 from umongo.data_proxy import data_proxy_factory
 from umongo import Document, EmbeddedDocument, Schema, EmbeddedSchema, fields, Reference
 from umongo.data_objects import List, Dict
+from umongo.exceptions import NotRegisteredDocumentError
 
 from .common import BaseTest
 
@@ -332,6 +333,12 @@ class TestFields(BaseTest):
             embedded_list_a = fields.ListField(
                 fields.EmbeddedField('EmbeddedDocument_A')
             )
+
+        # Can't make Marshmallow Schema before every EmbeddedField is resolved
+        with pytest.raises(NotRegisteredDocumentError):
+            schema = EmbeddedDocument_B.schema.as_marshmallow_schema()
+        with pytest.raises(NotRegisteredDocumentError):
+            schema = EmbeddedDocument_B.schema.fields['embedded_c'].as_marshmallow_field()
 
         @self.instance.register
         class EmbeddedDocument_C(EmbeddedDocument):
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -326,13 +326,24 @@ class TestFields(BaseTest):
             field = fields.IntField()
 
         @self.instance.register
-        class EmbeddedDocument_B(EmbeddedDocument):
+        class EmbeddedDocument_B(Document):
             embedded_a = fields.EmbeddedField('EmbeddedDocument_A')
             embedded_c = fields.EmbeddedField('EmbeddedDocument_C')
+            embedded_list_a = fields.ListField(
+                fields.EmbeddedField('EmbeddedDocument_A')
+            )
 
         @self.instance.register
         class EmbeddedDocument_C(EmbeddedDocument):
             field = fields.IntField()
+
+        @self.instance.register
+        class EmbeddedDocument_D(Document):
+            embedded_list_b = fields.ListField(
+                fields.EmbeddedField('EmbeddedDocument_B')
+            )
+            embedded_b = fields.EmbeddedField('EmbeddedDocument_B')
+
 
     def test_complexe_list(self):
 
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -319,6 +319,21 @@ class TestFields(BaseTest):
         repr_d = repr(d.get('list'))
         assert repr_d == "<object umongo.data_objects.List([2, 3, 4, 5])>"
 
+    def test_embedded_field(self):
+
+        @self.instance.register
+        class EmbeddedDocument_A(EmbeddedDocument):
+            field = fields.IntField()
+
+        @self.instance.register
+        class EmbeddedDocument_B(EmbeddedDocument):
+            embedded_a = fields.EmbeddedField('EmbeddedDocument_A')
+            embedded_c = fields.EmbeddedField('EmbeddedDocument_C')
+
+        @self.instance.register
+        class EmbeddedDocument_C(EmbeddedDocument):
+            field = fields.IntField()
+
     def test_complexe_list(self):
 
         @self.instance.register
--- a/umongo/builder.py
+++ b/umongo/builder.py
@@ -190,9 +190,9 @@ class BaseBuilder:
         field.instance = self.instance
         if isinstance(field, ListField):
             self._patch_field(field.container)
-        if isinstance(field, EmbeddedField):
-            for embedded_field in field.schema.fields.values():
-                self._patch_field(embedded_field)
+#        if isinstance(field, EmbeddedField):
+#            for embedded_field in field.schema.fields.values():
+#                self._patch_field(embedded_field)
 
     def _build_schema(self, template, schema_bases, schema_nmspc):
         # Recursively set the `instance` attribute to all fields
@@ -235,7 +235,8 @@ class BaseBuilder:
 
         # _build_document_opts cannot determine the indexes given we need to
         # visit the document's fields which weren't defined at this time
-        opts.indexes = _collect_indexes(nmspc.get('Meta'), schema.fields, bases)
+        #opts.indexes = _collect_indexes(nmspc.get('Meta'), schema.fields, bases)
+        opts.indexes_args = nmspc.get('Meta'), schema.fields, bases
 
         implementation = type(name, bases, nmspc)
         self._templates_lookup[template] = implementation
--- a/umongo/builder.py
+++ b/umongo/builder.py
@@ -11,7 +11,8 @@ from .embedded_document import (
 from .exceptions import DocumentDefinitionError, NotRegisteredDocumentError
 from .schema import Schema, on_need_add_id_field, add_child_field
 from .indexes import parse_index
-from .fields import ListField, EmbeddedField
+# from .fields import ListField, EmbeddedField
+from .fields import ListField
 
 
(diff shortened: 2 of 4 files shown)


lafrech:
I found two issues preventing the call of `fields.EmbeddedField('Score')` while `Score` is still undefined. It happens in the builders, when passing from template to instance.

#### Recursively pass instance through fields

This is what fails in traceback above. The builder tries to add a reference to `instance` to all fields recursively (into `ListField` and `EmbeddedField`). To loop on the fields inside the `EmbeddedField`, it needs to know the schema.

I managed to postpone this so that it is done only on the first access to `embedded_document_cls`, like this:

```py
    @property
    def embedded_document_cls(self):
        """
        Return the instance's :class:`umongo.embedded_document.EmbeddedDocumentImplementation`
        implementing the `embedded_document` attribute.
        """
        if not self._embedded_document_cls:

            self._embedded_document_cls = self.instance.retrieve_embedded_document(
                self.embedded_document)

            def _patch_field(instance, field):
                # Recursively set the `instance` attribute to all fields
                field.instance = instance
                if isinstance(field, ListField):
                    _patch_field(field.container)

            for embedded_field in self.schema.fields.values():
                _patch_field(self.instance, embedded_field)

        return self._embedded_document_cls
```

It might look hackish but it seems to work.

#### Indexes collection

But there's another problem. The document builder then fails at `_collect_indexes`. At registration time, the builder wants to get the indexes, and for that, it needs to inspect the fields recursively in the schema (the recursion occurs in `map_to_fields` in the `EmbeddedField`).

A way around this could be to postpone indexes collection (lazyness again) to their first use. This could be achieved more or less the same way: make `indexes` a `property` of `opts` and resolve it the first time it is called:

```py
    @property
    def indexes(self):
        if not self._indexes:
            # Insert collect indexes logic here
            # ...
        return self._indexes
```

except it is a bit more complicated here because the collect function needs values known at registration time by the `Builder` instance that we must store in the `DocumentOpts` instance until the first call to `opts.indexes`.

The new commit in this PR tries to do that quick and dirty just to see if it would work. Tests pass...

#### `EmbeddedField.as_marshmallow_field()`

There might be a third issue, though.

`EmbeddedDocument`'s `as_marshmallow_field` method uses  `self._embedded_document_cls`.

I'm not sure why you used `self._embedded_document_cls` here and not `self.embedded_document_cls`.

I changed it to `self.embedded_document_cls`, but the problem remains.

If it is called before the `EmbeddedDocument` is resolved, it will fail (see third commit).

I don't see any way around that. You can't make the schema before the EmbeddedDocument is created. Well unless we manage to also postpone the schema creation, pass a string to `ma_fields.Nested` in `EmbeddedField.as_marshmallow_field()`, and later register the schema in Marshmallow's class_registry ([which is forbidden as it is private API](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/class_registry.py#L8)). This doesn't sound good.

Maybe it is an acceptable limitation.

I'm a bit worried about it. We organize our flask app in modules following the [divisional structure](http://exploreflask.com/en/latest/blueprints.html#divisional), each module containing model.py, resources.py and schema.py. Worries begun when needing (for instance) to embed an EmbeddedDocument from module A in a Document from module B and link a Document from B in a Document from A. That's when `EmbeddedDocument('String')` should kick in. This PR might solve this use case, but we also create the API schema in schema.py in each module so Marshmallow schema are created at module import time before all Document and EmbeddedDocument classes are created.

I'd rather not redesign the whole app following the [functional structure](http://exploreflask.com/en/latest/blueprints.html#functional-structure)...

It does not look trivial, so I don't expect an immediate answer (please surprise me !), but can you tell me already if you expect umongo to support this feature (I mean `fields.EmbeddedField('Score')`, not the `as_marshmallow_schema` stuff)?

(It's a pity flake issues prevent the other tests from running.)

lafrech:
I just tested this with my app for further testing.

I can confirm those "fixes" allow me to cross-embed `EmbeddedDocument` across models from different modules imported sequentially, thanks to lazy resolution. And more trivially to manage the "`EmbeddedDocument` declared later" case.

I'm still in trouble when creating API schemas for module A if it embeds an `EmbeddedDocument` from module B and module B loads after module A, but I think I'm on my own here.

There has to be a limit to lazyness (trust me, I've been trying hard to reach it, never did !). umongo is not responsible for what I do with the schemas it creates. So if I decide to add further treatment in schema.py in each module at import time, this is my own problem.

Currently, for instance each module registers its API schemas in the API spec at import time, so the whole app is already broken if I have schemas for cross-embedded modules: module A registers its schemas, it has to load module B, which in turn has to load module A. umongo's `as_marshmallow_schema` adds nothing to the problem.

This is a design issue on my side. Maybe using functional structure would solve this. Maybe it would raise other issues. I'm interested by feedback from anyone courageous enough to have read all this gibberish.

#### TL; DR

- The `as_marshmallow_schema` issue is on user side, not umongo's. Pretend I didn't mention it.

- The fixes in the PR do solve the issue. They need serious rework, though. Not to mention units tests. Do you agree on the principle?

talwrii:
I really wanted this feature today (for recursive embedded documents).

talwrii:
Hmm, I'm keen to get something like this implemented.

If I get this rebased against the version 3 beta and fix any errors do you think I might be able to get this into the 3.0 release?

lafrech:
Great @talwrii.

This thread is old and I didn't go through it again. I think it needed not only tests but also a serious rework. Not sure whether it is easier to rebase or start from scratch. But yes, I'd be happy to get help solving this, so if you think you can do it, that's fantastic.

And I'm happy to have umongo 3 beta testers. My own app code doesn't use all features, so it's nice to have it tested on other codebases.

lafrech:
Update.

**Recursively pass instance through fields**

I removed the embedded document case as it was useless (#308) so this is not an issue anymore.

**Indexes collection**

Still an issue.

But since I removed the use of opts.indexes in `DuplicateKeyError` management (#309), those collected indexes are only used (internally) in `ensure_indexes`.

A plan could be to postpone indexes collection by making it manual through a method in `Instance`. That method could be called automatically in `ensure_indexes`.

And then this issue would be fixed.

lafrech:
We can add a `collect_indexes` method to DocumentImplementation, or even make it a cached property.

But since the method access the parent class indexes it must either run in the right order (hard to enforce if called manually) or be recursive (a cached property would fit the recursive requirement).

Since the doc would be registered, all its parents would be registered as well so their indexes would be collectable already.

I see no blocker here.

I begun drafting something.

I noted a test broken because a wrongly defined embedded doc wouldn't raise on registration due to collect_indexes not being called. This is a side effect. We may have to fix or ignore.

Issue #71, opened 2016-11-22
ListField not validated on append


lafrech:
I'm not sure whether this is expected behavior or not.

One can set a validator on a `ListField`, but it is used only when setting values, not when appending.

Here's a trivial example:

```py
        @db_instance.register
        class Doc(Document):
            l = fields.ListField(
                fields.IntField,
                validate=NoneOf(([1, 2, 3],))
            )

        d = Doc()

        with pytest.raises(ValidationError):
            d.l = [1, 2, 3]

        d.l = [1, 2]

        # The test fails here. No Exception raised.
        with pytest.raises(ValidationError):
            d.l.append(3)

        # However, an exception is raised here.
        with pytest.raises(ValidationError):
            d.l = d.l
```

This does not seem consistent. Assuming this is a bug/shortcoming, I'm not sure how I would address this.

My use case is a custom validator that checks a subitem is unique in a list of embedded documents: I have a ListField of EmbeddedField and I want to be sure that there is not two embedded documents with the same value for some specific field (or list of fields).

The validator is applied to the list, as it can obviously not be applied to the embedded documents individually.

Worse, I need it to be applied not only when an embedded document is appended to the list, but also when an embedded document in the list is mutated. If the latter is not feasible, I guess I can work this around by making the field read_only in my API. Better than nothing.

lafrech:
Unless I'm missing something, this is a really blocking issue, as this allows to set invalid data that will not be checked even at commit time.

I could modify `List` to keep a reference to its parent field and call

```py
    self._parent_field._validate(self)
```

in `set_modified` to ensure the field is validated each time the list is modified.

However, this is not satisfying because

- The validation error will be raised after the data is modified, so if the error is caught, the data is still wrong.

- This does not address the mutation of a `List` element. Not sure how this could be addressed.

Or should we live with it? In this case, it would be nice to at least validate the data on commit. I dug a bit into this and I'm afraid it's not easier.

What I can do is add pre_insert / pre_update methods in `Doc` where I manually validate the values:

```py
self.Schema().fields['l'].validate(self.l)
```
Is there a better workaround for this?

touilleMan:
You're totally right, this is a serious issue.

> Worse, I need it to be applied not only when an embedded document is appended to the list, but also when an embedded document in the list is mutated. If the latter is not feasible, 

This should be feasible with umongo !

> I could modify List to keep a reference to its parent field and call

[...]

PR #80, opened 2016-12-08
Can't copy/deepcopy a Document or EmbeddedDocument

lafrech:
I just noticed `Document` and `EmbeddedDocument` break `copy.copy` and `copy.deepcopy`.

Failing tests attached.

### DIFF
--- a/tests/test_document.py
+++ b/tests/test_document.py
@@ -1,5 +1,6 @@
 import pytest
 from datetime import datetime
+from copy import copy, deepcopy
 from bson import ObjectId, DBRef
 
 from umongo import (Document, EmbeddedDocument, Schema, fields, exceptions,
@@ -331,6 +332,14 @@ class TestDocument(BaseTest):
 
         with pytest.raises(NotImplementedError):
             Doc()
+
+    def test_copy(self):
+        john = self.Student(name='John Doe', birthday=datetime(1995, 12, 12), gpa=3.0)
+        copy(john)
+
+    def test_deepcopy(self):
+        john = self.Student(name='John Doe', birthday=datetime(1995, 12, 12), gpa=3.0)
+        deepcopy(john)
 
 
 class TestConfig(BaseTest):
--- a/tests/test_embedded_document.py
+++ b/tests/test_embedded_document.py
@@ -1,4 +1,5 @@
 import pytest
+from copy import copy, deepcopy
 from marshmallow import ValidationError
 
 from umongo.data_proxy import data_proxy_factory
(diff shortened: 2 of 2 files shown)


touilleMan:
https://docs.python.org/3.6/library/copy.html#module-copy

We could use `__copy__()` and `__deepcopy__()` special methods to solve this trouble. I'm not sure whether we should allow copy/deepcopy or not given they can bring subtle errors quickly (for example shallow copy of a document with embedded fields, then modifying the embedded field)

So want we can do for the moment is just raise a `RuntimeError` explaining copy and deepcopy are not allowed. And wait to have a better view (for example fixing #71 first) before doing more.

lafrech:
> We could use `__copy__()` and `__deepcopy__()` special methods to solve this trouble.

From what I've read, yes, we may use those to specify the behaviour we expect. Sometimes, people use them to fix what they have broken when playing with `__getattr__()`, while they should be fixing their mess instead, but in umongo, there may be things that need to be addressed in dedicated `__copy__` and `__deepcopy__` methods anyway.

>  I'm not sure whether we should allow copy/deepcopy or not given they can bring subtle errors quickly

I stumbled upon this while trying to deepcopy an `EmbeddedDocument` with several nesting levels. I ended up crating a `copy()` method at each level that in fact returns a deep copy, which is what I want. Supporting deepcopy would have avoided doing this in the model code.

I don't think the concept of copy/deepcopy is problematic. It should be pretty clear to anyone that an `EmbeddedDocument` is mutable and not duplicated when shallow copying parent Doc/EmbeddedDoc.

However, regarding `Document`, I'm not sure what the user would expect. If you copy a Document, you also copy the ID, but since you're copying it, you'll probably want a different ID, otherwise you'd be working on the original instance. Maybe the copy should return a Document with the same fields except for the ID field (and `is_created` set to False, see https://github.com/Scille/umongo/issues/83). But doing so, are we beginning to walk a dangerous path, making too many assumptions?

Overall, I'd say:

- copy / deepcopy should behave like usual copy / deepcopy: deepcopy duplicates everything while copy does not duplicate nested structures.

- I don't see anything specific to do for `EmbeddedDocument` (but should we add `__copy__`/`__deepcopy__` methods or is today's error fixable in the first place?)

- [deep|]copying a doc could return a different doc, with no id (unless the doc has an explicit id field) and `is_created` set to False, as if the doc had just been instantiated from scratch with the same values.

> So what we can do for the moment is just raise a RuntimeError explaining copy and deepcopy are not allowed. And wait to have a better view (for example fixing #71 first) before doing more.

Alright.

lafrech:
Closing in favor of https://github.com/Scille/umongo/pull/157.

Issue #81, opened 2016-12-11
Auto-indexes in Instance?


lafrech:
What would you think about a kind of semi-automatic opt-in indexes generation like this?

```py
# Mark Document for indexing
@instance.register(auto_index=True)
class MyDocument(Document):
...
```
In Instance:

```py
    def register(self, template, as_attribute=True, auto_index=False):
        # Retrieve the template if another implementation has been provided instead
        template = get_template(template)
        if issubclass(template, DocumentTemplate):
            implementation = self._register_doc(template)
+           if auto_index:
+               self._auto_indexes.append(implementation)
        else:  # EmbeddedDocumentTemplate
+           if auto_index:
+               raise SomeException
            implementation = self._register_embedded_doc(template)
        if as_attribute:
            setattr(self, implementation.__name__, implementation)
        return implementation

+   def ensure_indexes(self):
+       for doc in self._auto_indexes:
+           doc.ensure_indexes()
```

lafrech:
I'm working on a better approach

- using `Meta` in `Document` to specify `auto_indexes`
- using an `auto_indexes` attribute in `LazyInstance` to force `auto_indexes` on all `Document`s at `init`

touilleMan:
`ensure_indexes` has been dropped by pymongo due to potential concurrency issues when dropping & recreating rapidly the database. This leave us only with `create_indexes` which doesn't internally hold a timer, meaning each request should send a `create_indexes` first before actually send the request.

This is currently what happens with [mongoengine + pymongo>=3](https://github.com/MongoEngine/mongoengine/issues/1446), so I think requiring user to explicitly create the indexes makes him to to integrated it into workflow instead of relying on half broken magic.

PR #82, opened 2016-12-12
Add auto_indexes feature to LazyInstance

lafrech:
Closes https://github.com/Scille/umongo/issues/81.

What do you think about this ?

It works only on documents that are already registered when calling `instance.init()`.

TODO:

- [ ] Add tests for drivers other than PyMongo (currently only PyMongo is tested)

- [ ] Document the feature

- [ ] Also ensure indexes on documents registered after `instance.init()`?

### DIFF
--- a/tests/frameworks/test_pymongo.py
+++ b/tests/frameworks/test_pymongo.py
@@ -9,6 +9,7 @@ from ..fixtures import classroom_model, instance
 from umongo import (Document, EmbeddedDocument, fields, exceptions, Reference,
                     Instance, PyMongoInstance, NoDBDefinedError)
 
+from umongo.builder import camel_to_snake
 
 # Check if the required dependancies are met to run this driver's tests
 major, minor, _ = get_pymongo_version()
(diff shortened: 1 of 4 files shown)


lafrech:
> Obviously, it works only on documents that are already registered when calling instance.init().

I think this could be addressed in `Instance._register_doc`.

@touilleMan , waiting for your feedback before going any further. This is not critical to us.

touilleMan:
I think this feature is pretty interesting given calling `ensure_index` by hand is cumbersome and error prone ;-)

Few remarks:

1) With not implementing this for `Instance` as well as `LazyLoaderInstance` ?

2) The trouble with implementing this at the generic instance level is `ensure_indexes` won't do the same thing depending of the framework used: the indexes will be up and ready for pymongo were for asyncio and twisted only coroutine has been registered (which will be executed in the future... maybe... or can raise a exception which won't be catched...).
So I would have this feature in the builder given each framework has it own implementation:

```python
class Instance:
    ...
    def register(self, template):
        # overload the default register to add a synchronous creation of indexes if needed
        implementation = super().register(template)
        if self.auto_indexes:
            self.builder.ensure_indexes(implementation)

class LazyLoaderInstance:
    ...
    def init(self, db):
        ...
        # Now we can finish initialization of all the documents
        for doc in self._doc_lookup.values():
            self.builder.ensure_indexes(doc)

class PyMongoBuilder(BaseBuilder):

    BASE_DOCUMENT_CLS = PyMongoDocument

    ...

    def ensure_indexes(self, implementation):
        # For async framework, we create a reactor just for this coroutine so 
        # we end up with something synchronous anyway
        implementation.ensure_indexes()    
```

3) I don't like putting the `auto_indexes` option into the document's meta fields: I feel like document declaration should be independent of it execution as much as possible (we are defining a `Template` after all !). Using `auto_indexes` or not depend of which instance is used and not on what is the document made of.
I think it would be better to add an option to the `instance.register` instead:

```python
instance = Instance(db, auto_indexes=False)

@instance.register(force_auto_indexes_value=True)
class MyDoc(Document):
    ...
```

lafrech:
Thank you for your feedback.

> 1. Why not implement this for Instance as well as LazyLoaderInstance ?

I don't remember why I did not do it for `Instance`. I think I had a reason, probably a bad one...

> 2. & 3.

Good points. 

`register(force_auto_indexes_value=True)` was my first intent in https://github.com/Scille/umongo/issues/81, but then I changed my mind. I shouldn't have...

touilleMan:
I'm closing this because pymongo's ensure_index [has been deprecated](http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.ensure_index).

This makes creating automatically indexes much more complicated (plenty of complex concurrency bugs there, that's the reason pymongo dropped the feature in the first place).

For instance, this "don't bother user with this" approach created serious [perf issues with mongoengine](https://github.com/MongoEngine/mongoengine/issues/1446) that are now complicated to fix.

PR #84, opened 2016-12-22
Fix inherit embedded documents

patlach42:


lafrech:
This seems to solve my `EmbeddedDocument` issues.

It makes my failing test cases pass: https://github.com/Scille/umongo/pull/67.

I still don't get why we're calling `_deserialize_from_mongo` when not deserializing from Mongo, but this is the pattern used in the other fields, so it is probably the way to go, rather than the PR I sent (https://github.com/Scille/umongo/pull/56).

Back to the office (2017), I'll check that in our application tests.

Thanks @patlach42!

touilleMan:
Thanks a lot @patlach42 for the patch :smiley: 

@lafrech considering the use of `_deserialize_from_mongo` inside `_deserialize`, the idea is that we use first use vanilla marshmallow fields to do the deserialization (for example converting `"507f1f77bcf86cd799439011"` into `ObjectId("507f1f77bcf86cd799439011")`), then we end up with what pymongo would give us back from the database.
From there we use `_deserialize_from_mongo` to create the OO world object (`ObjectId("507f1f77bcf86cd799439011")` into the `umongo.data_objects.Reference()`).

This works great for Reference field, but not that much for recursive fields as you show here (List, Dict and Embedded). I've added some comments in this regard and try to limit the use of `_deserialize_from_mongo` in `_deserialize` only in `EmbeddedDocumentField` (couldn't find an elegant way not to use it for the moment)

@patlach42 I've added you to the AUTHORS.rst, notify me if you want to update those info :+1: 
@lafrech can you make sure the issues/PR you opened are fixed by this patch, then close them

lafrech:
@touilleMan yes, I'll make sure to do that (next year). I'm pretty sure the case is closed.

@imbolc is missing from AUTHORS.rst as well.

touilleMan:
> @imbolc is missing from AUTHORS.rst as well.

Done :smile:

lafrech:
@touilleMan, would you mind releasing a `v0.13.0`? Or may I?

(This was our last blocking issue. Thanks to this fix, our app should be able to use umongo master instead of a patch branch!)

touilleMan:
@lafrech today will be your first umongo release ^^

Check list:
- update `HISTORY.rst` to rename the `dev` entry with new version name
- use `bumpversion` tool to generate the release (e.g. `bumpversion minor` to go from 0.12.0 to 0.13.0), this should create a new commit and a tag named `v0.13.0`
- push to github both branches AND tags (by default only branches a pushed) `git push && git push --tags`

Once untitests have been run&passed on travis, the tag will be automatically send to pypi \o/

lafrech:
Great!

Done.

@patlach42, your fix is released in v0.13.0.

Issue #91, opened 2017-02-03
Synchronizing referencer and referenced document


lafrech:
I've been hit by this. Sometimes you have a Document and a referenced Document and when you're working on both instances at the same time you can find yourself in a sort of broken state.

I've pushed [a branch](https://github.com/Nobatek/umongo/commits/dev_jl_ref_fetch) with commented tests to explain the idea.

Currently, fetch() fetches from the database, so if you have a referenced Document instance you're working on, the changes won't be seen until it is committed and the Document is reloaded. This is shown [here](https://github.com/Nobatek/umongo/commit/852b58676fdc8e6f165de2d6e159668568fc631a).

The branch shows two measures that would mitigate this:

- Ensure that when assigning a Document to the reference field, that Document is set as _document so that fetch returns it even if it is not committed in DB. Shown [here](https://github.com/Nobatek/umongo/commit/b62fc9b756dd0deb6456d386f71a780f1811605b). Admittedly It looks like a bit of a hack. The point is to see what behavior this would provide.

- Provide a `force_reload` parameter to fetch() so that if the referenced Document instance is committed, it is possible to refresh the reference without reloading the whole Document. Shown [here](https://github.com/Nobatek/umongo/commit/7e0e46a9c92362b4a03fb178935e5deada1dea14).

I'm not sure about the first approach. It looks like it strives for providing a transparent reference management, but this is bound to fail at some point, so we'd rather keep things explicit.

On the other hand, the `force_reload` parameter is explicit and has limited impact. It just avoids reloading the whole Document (which could be problematic if work is ongoing). Without this (and without the other approach described above), there is no way to get fetch() to return the updated Document if it has changed in DB.

In fact, the more I think of it, the more the `force_reload` approach seems like the right approach to me. Anyway, I'm exposing both here for feedback.

lafrech:
Functionally, `force_reload` is equivalent to fetch() + reload() (shown [here](https://github.com/Nobatek/umongo/commit/d83649ca44cdaf6ab387260f02b7a2eff7072b01)).

Nothing really new, then. Yet, it might be handy as it allows one-liners since fetch returns the instance while reload does not.

touilleMan:
I like you suggestion of providing this `force_reload` option, this helps underlying the fact that de referenced document is cached and should be modified with care then.

Waiting for your PR now ;-)

lafrech:
Thanks for the feedback. We're awfully late on our app dev right now, so I can't promise I'll be able to do it any soon...

PR #96, opened 2017-03-03
Add force_reload to *Reference.fetch

lafrech:
Fixes https://github.com/Scille/umongo/issues/91.

(BTW, it looks like the `no_data` parameter has no effect. Is this a TODO?)

### DIFF
--- a/tests/frameworks/test_motor_asyncio.py
+++ b/tests/frameworks/test_motor_asyncio.py
@@ -275,6 +275,29 @@ class TestMotorAsyncio(BaseDBTest):
             assert isinstance(course.teacher, Reference)
             teacher_fetched = yield from course.teacher.fetch()
             assert teacher_fetched == teacher
+            # Change in referenced document is not seen until referenced
+            # document is committed and referencer is reloaded
+            teacher.name = 'Dr. Brown'
+            teacher_fetched = yield from course.teacher.fetch()
+            assert teacher_fetched.name == 'M. Strickland'
+            yield from teacher.commit()
+            teacher_fetched = yield from course.teacher.fetch()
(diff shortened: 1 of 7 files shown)


touilleMan:
> (BTW, it looks like the no_data parameter has no effect. Is this a TODO?)

Yep, seems like I put this as a placeholder for future optimization...

touilleMan:
merged, thx for the work 👍

lafrech:
Regarding `no_data` (or `exists`, I saw a comment in the code saying you'd like to rename it to `exists`), it would just need to add a projection to fetch only the `_id` field. What would you like it to return? A `Document` with only `id`? Or a boolean?

touilleMan:
I would say the return type of the function should not changed whether or not we provide `no_data`. So we should always return a `Document`, only passing the `no_data` param would mean all it fields but `_id` are missing (e.i. accessing them result in a `FieldNotLoadedError` exception)

lafrech:
Yes, that's what I thought, too.

lafrech:
What about adding a `projection` parameter to fetch? This would allow any projection possibility.

The `no_data` / `exists` case could be covered by either fetching with explicit `_id`-only projection, or we could introduce a `Reference.check_exists()` method that would do the projection and return a boolean (with `force_reload` to `True`?).

Even then, we'd still have a little performance penalty for the useless instantiation of the `Document`. To avoid this, `check_exists` could use the driver directly to avoid instantiation. I'm not sure this is really important.

touilleMan:
`projection` seems much better idea indeed ;-)

> To avoid this, check_exists could use the driver directly to avoid instantiation. I'm not sure this is really important.

This seems too low level for umongo (you want maximum perf, you do everything by hand using raw pymongo)

Issue #65, opened 2016-10-26
Pass meta attributes to schema in as_marshmallow_schema


lafrech:
Not sure whether this is a feature request or a question.

I would like to exclude a field from a schema of an `EmbeddedDocument` when calling `as_marshmallow_schema` on its container `Document`.

To do that on a Schema generated by `as_marshmallow_schema`, I would just call `as_marshmallow_schema` to get the Schema then subclass it to add a Meta class:

``` py
schema = Doc.as_marshmallow_schema()

class MySchema(schema):
    class Meta:
        exclude: ('hidden_field',)
```

It would be cool to be able to call 

``` py
schema = Doc.as_marshmallow_schema(meta={'exclude': ('hidden_field',)})
```

but it's not critical, at least in my current use of umongo.

It becomes more useful if the Schema I need to modify is nested in the document, so that the call to `as_marshmallow_schema` is not direct but cascaded through `EmbeddedField`. In this case, the meta attributes need to be provided to `as_marshmallow_schema` by `as_marshmallow_field` in 
`EmbeddedField`.

Maybe it could take the (absolutely untested) following form:

``` py
    # In BaseSchema
    def as_marshmallow_schema(self, params=None, base_schema_cls=MaSchema,
                              check_unknown_fields=True, missing_accessor=True,
                              mongo_world=False, meta=None):

        [...]
        if meta is not None:
            # Do what is to be done

    # In EmbeddedField
    def as_marshmallow_field(self, params=None, mongo_world=False):
        # Overwrite default `as_marshmallow_field` to handle nesting
        kwargs = self._extract_marshmallow_field_params(mongo_world)
        if params:
            nested_params = params.pop('params')
            nested_meta = params.pop('meta')
            kwargs.update(params)
        else:
            nested_params = None
            nested_meta = None
        nested_ma_schema = self._embedded_document_cls.schema.as_marshmallow_schema(
            params=nested_params, mongo_world=mongo_world, meta=meta)
        return ma_fields.Nested(nested_ma_schema, **kwargs)
```

Or am I missing an already existing feature?

touilleMan:
I like this very much ! It is simple to implement and really get rid of the boilerplates

Now I'm waiting for you PR ;-)

Issue #64, opened 2016-10-26
as_marshmallow_field params are not passed to the container in a ListField


lafrech:
In `ListField`, the `params` that are passed to `as_marshmallow_field` are used only for the underlying Marshmallow `List` but nothing is passed to the container.

https://github.com/Scille/umongo/blob/master/umongo/fields.py#L119

AFAIU, there is no way to pass `params` to the container, for instance an `EmbeddedField` inside the `ListField`. This sort of breaks the cascading mechanism.

touilleMan:
What about a special `container_params` field to define in `param` ?

lafrech:
Well, IIUC, we could do the same as EmbeddedField:

`params` is a dict with the parameters for the field, and it may include a `params` key for which the value is a dict with the parameters for the nested field.

This should work in all cases. As long as no field requires a parameter called `params`, since this would be a reserved word for nested field parameters.

Unless I missed something, this should be pretty easy.

I'll try to tackle this and https://github.com/Scille/umongo/issues/65. Not a priority right now, but I'll try to find the time.

lafrech:
Labelling bug as it is a flaw of the existing mechanism rather than a missing feature.

Assigning to myself. Doesn't mean I'll be doing this any soon. I'm on tight schedule and this is not blocking. If you find time to do that, don't refrain to do so!

PR #101, opened 2017-03-22
Pass as marshmallow schema params to nested schemas

lafrech:
Parameters of `as_marshmallow_schema` should be passed to nested Schemas.

In this implementation, I pass them to all fields, so that `EmbeddedField` gets it and pass it when calling `as_marshmallow_schema`.

Nothing really complicated here. I had to modify `as_marshmallow_field`'s signature, hence the modifications in unrelated fields.

Note that I changed `kwargs` into `field_kwargs` in those fields to make it obvious to the reader it is not the same `kwargs` we're talking about. It would work without the rename, but it would look like a mistake at first sight:

```python
    def as_marshmallow_field(self, params=None, mongo_world=False, **kwargs):
        # Oh, we're overwriting kwargs! Let's file a bug!
        kwargs = self._extract_marshmallow_field_params(mongo_world)
```

Another option was to use `_` but I'm not sure it is so common for kwargs:

```python
    def as_marshmallow_field(self, params=None, mongo_world=False, **_):
        kwargs = self._extract_marshmallow_field_params(mongo_world)
```

I preferred to rename the field kwargs inside the method:

```python
    def as_marshmallow_field(self, params=None, mongo_world=False, **kwargs):
        field_kwargs = self._extract_marshmallow_field_params(mongo_world)
```

I could rework this if you have a better idea.

This PR does _not_ address those two related issues I assigned myself a while ago: https://github.com/Scille/umongo/issues/64, https://github.com/Scille/umongo/issues/65...

### DIFF
--- a/umongo/abstract.py
+++ b/umongo/abstract.py
@@ -191,8 +191,7 @@ class BaseField(ma_fields.Field):
         params.update(self.metadata)
         return params
 
-    def as_marshmallow_field(self, params=None, base_schema_cls=MaSchema,
-                             check_unknown_fields=True, mongo_world=False):
+    def as_marshmallow_field(self, params=None, mongo_world=False, **kwargs):
         """
         Return a pure-marshmallow version of this field.
 
--- a/umongo/abstract.py
+++ b/umongo/abstract.py
@@ -43,8 +43,14 @@ class BaseSchema(MaSchema):
             instead of the OO world (default: False).
         """
         params = params or {}
-        nmspc = {name: field.as_marshmallow_field(params=params.get(name), mongo_world=mongo_world)
-                 for name, field in self.fields.items()}
+        nmspc = {
+            name: field.as_marshmallow_field(
+                params=params.get(name),
+                base_schema_cls=base_schema_cls,
+                check_unknown_fields=check_unknown_fields,
+                mongo_world=mongo_world)
+            for name, field in self.fields.items()
+        }
         name = 'Marshmallow%s' % type(self).__name__
         if check_unknown_fields:
             nmspc['_%s__check_unknown_fields' % name] = validates_schema(
@@ -185,7 +191,8 @@ class BaseField(ma_fields.Field):
         params.update(self.metadata)
         return params
 
-    def as_marshmallow_field(self, params=None, mongo_world=False):
+    def as_marshmallow_field(self, params=None, base_schema_cls=MaSchema,
+                             check_unknown_fields=True, mongo_world=False):
         """
         Return a pure-marshmallow version of this field.
 
@@ -194,14 +201,14 @@ class BaseField(ma_fields.Field):
         :param mongo_world: If True the field will work against the mongo world
             instead of the OO world (default: False)
         """
-        kwargs = self._extract_marshmallow_field_params(mongo_world)
+        field_kwargs = self._extract_marshmallow_field_params(mongo_world)
         if params:
-            kwargs.update(params)
+            field_kwargs.update(params)
         # Retrieve the marshmallow class we inherit from
         for m_class in type(self).mro():
             if (not issubclass(m_class, BaseField) and
                     issubclass(m_class, ma_fields.Field)):
-                m_field = m_class(**kwargs)
+                m_field = m_class(**field_kwargs)
                 # Add i18n support to the field
                 m_field.error_messages = I18nErrorDict(m_field.error_messages)
                 return m_field
--- a/tests/test_marshmallow.py
+++ b/tests/test_marshmallow.py
@@ -204,6 +204,33 @@ class TestMarshmallow(BaseTest):
         assert not ret.errors
         assert ret.data == bag.to_mongo()
 
+        # Check as_marshmallow_schema params (check_unknown_felds, base_schema_cls)
+        # are passed to nested schemas
+        data = {
+            'id': {'brief': 'sportbag', 'value': 100, 'name': 'Unknown'},
+            'content': [
+                {'brief': 'cellphone', 'value': 500, 'name': 'Unknown'},
+                {'brief': 'lighter', 'value': 2, 'name': 'Unknown'}]
+        }
+        ret = ma_schema.load(data)
+        assert ret.errors == {
+            'id': {'_schema': ['Unknown field name name.']},
+            'content': {
+                0: {'_schema': ['Unknown field name name.']},
+                1: {'_schema': ['Unknown field name name.']},
+            }}
+
+        ma_no_check_unknown_schema = Bag.schema.as_marshmallow_schema(check_unknown_fields=False)()
+        ret = ma_no_check_unknown_schema.load(data)
+        assert not ret.errors
+
+        class WithNameSchema(marshmallow.Schema):
+            name = marshmallow.fields.Str()
+
+        ma_custom_base_schema = Bag.schema.as_marshmallow_schema(base_schema_cls=WithNameSchema)()
+        ret = ma_custom_base_schema.load(data)
+        assert not ret.errors
+
     def test_marshmallow_bonus_fields(self):
         # Fields related to mongodb provided for marshmallow
         @self.instance.register
--- a/umongo/fields.py
+++ b/umongo/fields.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 
 from marshmallow import ValidationError, missing
-from marshmallow import fields as ma_fields, Schema as MaSchema
+from marshmallow import fields as ma_fields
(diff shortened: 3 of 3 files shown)


lafrech:
Better this way. I can squash if you want a cleaner history.

umongo/abstract.py
```diff
@@ -181,7 +184,7 @@ def _extract_marshmallow_field_params(self, mongo_world):
         params.update(self.metadata)
         return params
 
-    def as_marshmallow_field(self, params=None, mongo_world=False):
+    def as_marshmallow_field(self, params=None, mongo_world=False, **kwargs):
         """
         Return a pure-marshmallow version of this field.
```
touilleMan: You should also rename kwargs into field_kwargs here too.

touilleMan:
Given `as_marshmallow_schema` doesn't use `**kwargs` as params, I think we should explicitly pass them to `as_marshmallow_field` (i.e. add `base_schema_cls` and `check_unknown_fields` to `as_marshmallow_field` signature).

umongo/abstract.py
lafrech: I don't think so.

`field_kwargs` are passed in `params`.

`kwargs` are schema-level parameters, sort of. The implementation I propose passes them to all fields although they are only needed in `EmbeddedField` (to be passed to nested schema). We could pass them to `EmbeddedField` only (and through `ListField`) but I figured we might as well pass `kwargs` to all fields. 
There could be other shcema-level parameters that would affect all `as_marshmallow_fields` and this would be the right place for them.

lafrech:
> add base_schema_cls and check_unknown_fields to as_marshmallow_field signature

Look at the commits separately. What you suggest is what I did first. See the [second commit](https://github.com/Scille/umongo/pull/101/commits/6f332874b5b2aa4c4081020fc7cde23b9c2a9ccc), before the rework.

I reworked it to use `kwargs` instead because I thought it would look nicer. It makes it more explicit that those parameters may not be used in the field. It is easier in terms of maintenance, as well. Once `**kwargs` is added to the signature, you can add new arguments without breaking all the fields (and custom fields in client applications) that override `as_marshmallow_schema`.

> Given as_marshmallow_schema doesn't use **kwargs as params

OK, now I see what you don't like in my proposal: in `EmbeddedSchema.as_marshmallow_field`, we're passing all kwargs blindly without checking that they are in `as_marshmallow_schema`'s signature.

Maybe we can keep `kwargs` in `as_marshmallow_schema` and fix this in `EmbeddedField`:

```python
        schema_kwargs = []
        if hasattr(kwargs, 'base_schema_cls'):
            schema_kwargs['base_schema_cls'] = kwargs['base_schema_cls']
        if hasattr(kwargs, 'check_unknown_fields'):
            schema_kwargs['check_unknown_fields'] = kwargs['check_unknown_fields']
        nested_ma_schema = self._embedded_document_cls.schema.as_marshmallow_schema(
            params=nested_params, mongo_world=mongo_world, **schema_kwargs)
```

It looks a bit verbose, but at least it is located in only one place, not duplicated in all fields.

Or add `**kwargs` to `as_marshmallow_schema` but it sounds bad.

There may be other ways, functionally equivalent.

If think I get your reasons. The advantage of kwargs over explicit signature is the possibility to let this evolve without breaking existing fields (in umongo + custom fields in applications). But you may consider the whole mecanism stable enough to ignore this.

If you want to go explicit, I can revert the "rework" commit. Do you want me to do that?

lafrech:
I reworked this a bit.

I'm keeping kwargs in the signature of `as_marshmallow_field`, but I filter the kwargs that are passed to `as_marshmallow_schema` to avoid passing it any unknown argument. Like I suggested in previous comment but in a much less clumsy way.

We may get the best of the two solutions. Not too explicit, but still safe.

Also, not having `base_schema_cls` and `check_unknown_fields` in `EmbeddedField.as_marshmallow_field`'s signature allows not to duplicate their default values (`True` and `MaSchema`) from abstract.py. But I could put them in the signature if you preferred:

```python
    def as_marshmallow_field(self, params=None, mongo_world=False,
            base_schema_cls=MaSchema, check_unknown_fields=True, **kwargs):
        # Overwrite default `as_marshmallow_field` to handle nesting
        field_kwargs = self._extract_marshmallow_field_params(mongo_world)
        if params:
            nested_params = params.pop('params')
            field_kwargs.update(params)
        else:
            nested_params = None
        nested_ma_schema = self._embedded_document_cls.schema.as_marshmallow_schema(
            params=nested_params, mongo_world=mongo_world,
            base_schema_cls=base_schema_cls, check_unknown_fields=check_unknown_fields)
        return ma_fields.Nested(nested_ma_schema, **field_kwargs)
```

(I'm keeping kwargs in the signature for consistency with other fields: adding stuff here for any field doesn't break the others.)

lafrech:
This PR is the only patch left on my dev branch.

I think it works as expected and the only issue at stake here was about umongo API (whether to use **kwargs or not, etc.). I proposed several implementation (several commits in PR). If you tell me which one you prefer, I can cleanup and rebase. This would allow to close/merge https://github.com/Scille/umongo/pull/102 and https://github.com/Scille/umongo/pull/103 as well.

No rush on my side, except the more time passes, the more rebase conflicts and the more I need to struggle to remember what I did back then.

Just a bump as this is getting old, but no rush, really.

touilleMan:
@lafrech sorry for the time delay

As you may see, I don't have much time to allocate on umongo recently due to a demending day job and my open source contribution time being stretched between to many projects ;-)

Anyway, regarding this issue, I feel like it's been too long for me to have deep enough understanding. So I trust you on to choose the best implementation ;-)

(and don't hesitate to publish a new release once this is merged)

umongo/abstract.py
lafrech: Oh, you meant kwargs below, not in the signature. Yes, you're right.

lafrech:
I took it upon myself to merge this and two other PRs depending on it. This closes old issues. Great.

I also updated HISTORY.rst.

A release would be nice. I could try to take a stab at it if I find the procedure. Tell me if you want me to do it.

In any case, due to the newly merged changes, I'd rather have your approval before releasing.

touilleMan:
> A release would be nice. I could try to take a stab at it if I find the procedure. Tell me if you want me to do it.

Go on for the release.
- run bumpversion (https://pypi.org/project/bumpversion/), this should change version number in a couple of files (setup.cfg, setup.py and __version__.py) and create a new commit and tag
- push the master branch with the new commit, make sure CI is green
- push the tag, CI should do the deploy on pypi by itself
- have a beer, you earned it ;-)

Regarding the version number, maybe it's time to switch to v1.0.0 (the api seems stable enough, and then we could start doing real SemVer)

lafrech:
Good.

I'll have a look at the bugtracker to see if there's easy stuff to do before publishing 1.0.

- What about https://github.com/Scille/umongo/issues/152 (drop Python 3.4)? At least the "drop compatibility" part, not the async improvements.

- The [default/missing rework](https://github.com/Scille/umongo/pull/107) is a breaking change that could be put in 1.0 if it's ready.

lafrech:
Alright, let's release a v1.0.0. When marshmallow 3 is out and umongo is compatible, we'll release a v2 dropping Python 3.4. And when we have time, we can finalize the default/missing rework.

CI reports errors due to deployment being attempted to three times, once for each Python 3.6 build. The package was correctly deployed.

touilleMan:
> let's release a v1.0.0. When marshmallow 3 is out and umongo is compatible

But it seems you already released a v1.0.0 (cf d03391834b2784ee63a024c8fe7844069abf6918) do you mean a v2.0.0 ?

> we'll release a v2 dropping Python 3.4

You should replace the 3.4 by 3.5 in the `.travis.yml` (and add 3.7 also)

> CI reports errors due to deployment being attempted to three times, once for each Python 3.6 build. 

Yup, this should do the trick: https://docs.travis-ci.com/user/deployment/pypi/#upload-artifacts-only-once

lafrech:
> But it seems you already released a v1.0.0 (cf d033918) do you mean a v2.0.0 ?

I meant let's release a v1.0.0 (indeed I did), and postpone Python 3.4 removal and other breaking changes to next major (v2.0.0).

> You should replace the 3.4 by 3.5 in the .travis.yml (and add 3.7 also)

3.7 is a bit more tricky to achieve in Travis, but it is feasible (https://github.com/marshmallow-code/marshmallow/pull/858).

> Yup, this should do the trick: https://docs.travis-ci.com/user/deployment/pypi/#upload-artifacts-only-once

Yes.

PR #104, opened 2017-03-23
Fix equality tests

lafrech:
I've been having issues when comparing `EmbeddedDocument`s with nesting.

Sometimes, the nested field is not an `EmbeddedDocument` but `missing` or `None`, and the comparison fails raising `AttributeError` because `other._data` does not exist.

I tried to cover theses cases and maybe I'm covering a bit larger, but I don't think it should be an issue.


touilleMan:
👍

PR #106, opened 2017-05-04
Rework StrictDateTime

lafrech:
This commit

- factorizes the TZ awareness enforcement logic (I wonder why I didn't do that right away...)

- makes `StrictDateTime` bonus field reject `datetime.datetime` instances, which is coherent with `DateTime`. Accepting `datetime.datetime` instances is done in umongo's `StrictDateTimeField` only, just like with `DateTimeField`.

Should I had test for this?

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,8 @@ History
 
 dev
 ---
+* Fix StrictDateTime bonus field: _deserialize does not accept datetime.datetime
+  instances (see #106)
 * Add force_reload param to Reference.fetch (see #96)
 
 0.13.0 (2017-03-03)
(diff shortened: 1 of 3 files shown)


touilleMan:
Do you think tests are really useful for this (I mean if `StrictDateTime` is already well tested no need) ?
You should add a line about this in the `HISTORY.rst` however.

lafrech:
> Do you think tests are really useful for this (I mean if StrictDateTime is already well tested no need) ?

Nope. I could add a test for this change (test that it now rejects `datetime` instances), but I don't think it would be of much interest.

> You should add a line about this in the HISTORY.rst however.

Done !

touilleMan:
merged :+1:

PR #109, opened 2017-05-31
Fix deletion of field proxied by a property

lafrech:
Like `__setattr__`, `__delattr__` should only get to the DataProxy for fields, not other attributes. Otherwise, when a field is proxied by a property, its deletion fails.

### DIFF
--- a/tests/test_document.py
+++ b/tests/test_document.py
@@ -117,6 +117,29 @@ class TestDocument(BaseTest):
         with pytest.raises(KeyError):
             del john['missing']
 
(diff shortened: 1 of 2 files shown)


lafrech:
Note we could factorize that part

```python
        if not self.__real_attributes:
            type(self).__real_attributes = dir(self)
```

for instance by making `__real_attributes` a property .

Since it's only two lines I assumed it was fine to keep it duplicated.

touilleMan:
good job !

PR #112, opened 2017-06-08
Remove EmbeddedDocument._modified

lafrech:
Looks like this is never use. `is_modified` only checks the data proxy.


touilleMan:
Old dead code, thanks for pointing out 👍

PR #121, opened 2017-10-13
Raise exception with Document name when unknown field found in DB

lafrech:
Currently, when an unknown field is found in database using a strict DataProxy, the KeyError is not caught. When dealing with deeply nested documents, it can be a bit cumbersome to track the faulty EmbeddedDocument.

This PR creates a dedicated exception that prints the name of the data proxy to ease the debug procedure.


lafrech:
Perhaps `UnknownFieldInDBError` could inheritate from both `UMongoError` and `KeyError` so that it would still be caught by a `try / except KeyError`, both for compatibility with existing code and for semantic, after all, it is some sort of `KeyError`. Feedback welcome.

lafrech:
I can update the translation files once you agree with this change.

touilleMan:
Thanks for the PR (and sorry for the delay of the reply ^^)
This seems good for me, I don't think the exception should inherit `KeyError` (as you said it would be more a compatibility thing than really something wanted, so we can just release a new major version and begone with it)

So just update the Changelog and the translation file and I'll merge and release a new version :+1:

lafrech:
I'm sorry, I can't figure out how to update `messages.pot`.

Is there a command line documented somewhere?

I know there are different ways of doing this and generally, you set options in a config file so that you only have a simple command to run. I've been looking in setup.py, setup.cfg, Makefile,... and didn't find anything. Did I miss it anything?

 I think we need to pass specific arguments for `_N` to be understood, for instance.

I should also update the FR translation in the example and recompile the `.mo` file. Help welcome here as well.

touilleMan:
This i18n stuff is quiet complex, what a relief I kept the commands in a file somewhere ;-)

```
pybabel update -i ../../messages.pot -l fr -o translations/fr/LC_MESSAGES/messages.po
pybabel extract umongo -o messages_v2.pot
```

I think the first one is to regenerate the .po once you have updated the .pot (should be launched from the `examples/[flask|klein|muffin]` directory
Second one regenerate the .pot by extracting strings from the project. Note the file is called with a `_v2` not to overwrite the old translation. This is because we have to edit the translation by hand to add the marshmallow strings entries.

Given the pain I had to retrieve this, do you thing you could document this process in the i18n part of the doc ?

touilleMan:
can you update this branch ?

lafrech:
Oops, right, sorry.

Rebased!

touilleMan:
:taco:

enugentdt:
So, I have an interesting problem, that comes from this pull request & code. I am using this in an environment where I may add fields to collections in the db. However, I will have multiple servers, some running old code, some running new. If I make a change to the db, and push some new code, the servers with old code will all crash. The solution roughly is to just ignore unknown fields. Would it make sense to add a flag to not raise an UnknownFieldInDBError? I don't know if what I'm saying actually makes sense, so I'm happy to clarify if need be. I'm also happy to make a PR to fix this, if it would help.

Thanks!

lafrech:
This PR only adds an explicit error where the code would crash anyway with an error that is harder to rootcause.

If you want to pass unknown fields, use the `strict=False` `Meta` option.

```py
class MyDoc(Document):
    class Meta:
        strict = False
```

I don't think there's a way to _ignore_ those fields. At least, this option prevents the crash.

enugentdt:
Excellent! That works perfectly. Thanks so much!

PR #124, opened 2017-10-23
Document translation process

lafrech:
I added a few helpers in the Makefile and documented the translation process.

Since this is the process to translation umongo itself as well as the flask example files, it applies when modifying umongo, so I documented it in CONTRIBUTING.rst. People using umongo may have different strategies and can take inspiration in this for their process, but I didn't think this fitted in the user guide (although I admit a few hints in there could be nice as well).


touilleMan:
coool ^^

PR #133, opened 2018-07-23
fix error due to typo in flask example

mandarup:
- fixes `TypeError: as_marshmallow_schema() missing 1 required positional argument: 'self'`

### DIFF
--- a/examples/flask/app.py
+++ b/examples/flask/app.py
@@ -130,7 +130,7 @@ def update_user(nick_or_id):
     if not user:
         abort(404)
     # Define a custom schema from the default one to ignore read-only fields
-    UserUpdateSchema = User.Schema.as_marshmallow_schema(params={
+    UserUpdateSchema = User.schema.as_marshmallow_schema(params={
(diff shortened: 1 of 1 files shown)


touilleMan:
thanks, I've manually merged both of your PRs :+1:

mandarup:
thanks @touilleMan. I wasn't sure how else I could reach you .. Do you have any idea if this project is going to be maintained in foreseeable future?  I am starting out with a project and I am struggling to choose between ODM options. any recommendations/advice?

touilleMan:
@mandarup I've end up creating umongo because mongoengine really was too broken to be fixed, so I obviously don't recommend you this one (I've no experience with other ODM, but they all seemed to be much more SQL inspired ORM than real mongo ODM at the time I was looking for an alternative to mongoengine)
Considering umongo, I'm not planning any big evolution on it, but I'm really open to PR and fixes ;-)

mandarup:
@touilleMan Thanks for your response, it's very valuable to me, much appreciated!!

PR #147, opened 2018-11-23
Fix CI tests

lafrech:
- Update flake8 config to ignore W504 (see rationale in https://github.com/marshmallow-code/apispec/pull/322).
- Don't call fixtures directly, which is forbidden with pytest 4.
- Don't test index version. The index version depends on the database version (and configuration). It should not be tested here. This breaks the CI tests as they are now run on MongoDB 3.4 which default to v2.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,7 +15,7 @@ replace = __version__ = '{new_version}'
 universal = 1
 
 [flake8]
-ignore = E127,E128
+ignore = E127,E128,W504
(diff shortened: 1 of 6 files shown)


touilleMan:
What's this `<CursorHold>` file ? 

Otherwise seems good to me :+1:

lafrech:
Oops, the `<CursorHold>` file was generated by vi and committed accidentally. I don't know what it is for. Removed.

PR #149, opened 2018-11-28
Update to pymongo 3.7

lafrech:
https://api.mongodb.com/python/current/changelog.html

> Deprecated `pymongo.collection.Collection.count()` and `pymongo.cursor.Cursor.count()`. These two methods use the count command and may or may not be accurate, depending on the options used and connected MongoDB topology. Use `count_documents()` instead.

This PR is based on #148.

It adds a `count` method to `PyMongoDocument`. (I named it `count` and not `count_documents` to match umongo's `TxMongoDocument`. I have no objection to naming it `count_documents`.)

I'd do the same for `MotorAsyncIODocument` except it requires motor>=2.0, which is not supported.

Since we already require pymongo 3.x, I figured bumping the requirement wouldn't be too much of an issue and would save us the pain of supporting both 3.7 and 3.6-. I could rework to only add the new `count` method if using 3.7 and only require 3.7 in the tests. But I'm not sure it is worth the trouble.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -75,12 +75,11 @@ Quick example
 
 Get it now::
 
-    $ pip install umongo
-    $ pip install my-mongo-driver  # Note you have to manually install the mongodb driver
+    $ pip install umongo           # This installs umongo with pymongo
+    $ pip install my-mongo-driver  # Other MongoDB drivers must be installed manually
 
 Or to get it along with the MongoDB driver you're planing to use::
 
-    $ pip install umongo[pymongo]   # choose
-    $ pip install umongo[motor]     # one
-    $ pip install umongo[txmongo]   # of
-    $ pip install umongo[mongomock] # them ;-)
+    $ pip install umongo[motor]
+    $ pip install umongo[txmongo]
+    $ pip install umongo[mongomock]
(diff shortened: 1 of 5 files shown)


touilleMan:
> It adds a count method to PyMongoDocument. (I named it count and not count_documents to match umongo's TxMongoDocument. I have no objection to naming it count_documents.)

I'd better keep `count_documents` if it's the name in the pymongo driver (I guess people are more interested in consistency between umongo and the driver they use than between drivers)
Another possibility would be to provide both name (`count` and `count_documents`), as you like.

> Since we already require pymongo 3.x, I figured bumping the requirement wouldn't be too much of an issue and would save us the pain of supporting both 3.7 and 3.6

Agree, let's keep things simple ;-)

touilleMan:
otherwise seems good to me :+1:

lafrech:
Alright.

I changed to `count_documents`. Note that my implementation differs from pymongo's in that the filter parameter is optional and defaults to `{}`. I can change that if you prefer to adhere more strictly. I don't understand why pymongo didn't make this argument optional. Maybe their good reason would apply here, too.

PR #153, opened 2018-11-30
Add skip_existing parameter to Travis CI deployment

lafrech:


lafrech:
Let's merge this.

This flag is documented here: https://docs.travis-ci.com/user/deployment/pypi/#upload-artifacts-only-once:

> By default, Travis CI runs the deploy stage for each python and environment that you specify. Many of these will generate competing build artifacts that will fail to upload to pypi with a message something like this:
> 
>     HTTPError: 400 Client Error: File already exists. See https://pypi.org/help/#file-name-reuse for url: https://upload.pypi.org/legacy/
> 
> To avoid this, use the skip_existing flag:
> 
>     deploy:
>       provider: pypi
>       user: ...
>       password: ...
>       skip_existing: true

See https://github.com/Scille/umongo/pull/101#issuecomment-443165820.

Issue #129, opened 2018-01-12
Port to Marshmallow 3


lafrech:
Creating this issue to discuss things related to Marshmallow 3.

Marshmallow 3 will introduce breaking changes. Assuming we want to keep compatibility with Marshmallow 2, we'd need to introduce conditional code to adapt to both versions and test on both v2 and v3. This can be started before Marshmallow 3.0.0 is out.

Regarding the tests, here's how I did it on another lib: https://github.com/Nobatek/flask-rest-api/commit/cd23fa2ccfba47b51ec776ff47c74b447e02b378. Inspired from `mongoengine` testing on several `pymongo` versions. Obviously this means the test will break each time the dev branch introduces breaking changes.

Another approach would be to work on a separate branch and only merge when Marshmallow 3 is stable.

In the meantime, we can also help closing v3 issues in Marshmallow to speed up the process: https://github.com/marshmallow-code/marshmallow/issues?q=is%3Aopen+is%3Aissue+milestone%3A3.0.

touilleMan:
What kind of changes introduce marshmallow 3 ?

> Assuming we want to keep compatibility with Marshmallow 2

Not sure if it's something wanted as compatibility means more work...

It obviously depend on how much work is involved, but I say we could release a new major version with marshmallow 3 and from there backport really important patches to the actual major version (which would become the marshmallow 2 support branch)

lafrech:
> What kind of changes introduce marshmallow 3 ?

See [changelog](https://github.com/marshmallow-code/marshmallow/blob/dev/CHANGELOG.rst) and [upgrading to 3.0](https://github.com/marshmallow-code/marshmallow/blob/dev/docs/upgrading.rst) for what is done already. And [issues milestones 3.0](https://github.com/marshmallow-code/marshmallow/issues?q=is%3Aopen+is%3Aissue+milestone%3A3.0) for what is to come.

All schemas are now strict, and only data is returned. Errors trigger an exception containing error messages. This and the other changes should be easy fixes in umongo. I even hope this can be done with a single umongo version supporting both marshmallow versions, with a if `MARSHMALLOW_2` test in a few places. This would be less work than backporting patches IMHO.

[...]

PR #154, opened 2018-12-03
Marshmallow 3 compatibility

lafrech:
Closes https://github.com/Scille/umongo/issues/129.

The failing test is due to an issue in marshmallow that was fixed since 3.0.0rc1 was released. It'll be fixed with next marshmallow release.

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -517,16 +517,14 @@ by subclassing it:
     >>> patch_dog_schema.load({'name': 'Scruffy', 'breed': 'Golden retriever'}).errors
     {'_schema': ['Unknown field name breed.']}
     >>> ret = patch_dog_schema.load({'name': 'Scruffy'})
-    >>> ret.errors
-    {}
-    >>> ret.data
+    >>> ret
     {'name': 'Scruffy'}
 
 Finally we can integrated the validated data into OO world:
 
 .. code-block:: python
 
-    >>> my_dog.update(ret.data)
+    >>> my_dog.update(ret)
     >>> my_dog.name
     'Scruffy'
 
@@ -638,9 +636,9 @@ So when you use ``as_marshmallow_field``, the resulting marshmallow field's
         skill = fields.StrField(default='Dummy', marshmallow_default=missing)
 
     ret = Employee.schema.as_marshmallow_schema()().load({})
-    assert ret.data == {'name': 'John Doe', 'birthday': datetime(2000, 1, 1, 0, 0, tzinfo=tzutc()), 'skill': 'Dummy'}
+    assert ret == {'name': 'John Doe', 'birthday': datetime(2000, 1, 1, 0, 0, tzinfo=tzutc()), 'skill': 'Dummy'}
     ret = Employee.schema.as_marshmallow_schema()().dump({})
-    assert ret.data == {'name': 'John Doe', 'birthday': '2000-01-01T00:00:00+00:00'}  # Note `skill` hasn't been serialized
+    assert ret == {'name': 'John Doe', 'birthday': '2000-01-01T00:00:00+00:00'}  # Note `skill` hasn't been serialized
 
 
 Field validate & io_validate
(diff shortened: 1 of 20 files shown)


lafrech:
Updated.

Many failing tests because this is based on latest unreleased commits from dev branch. This will be fixed with next marshmallow release.

wfscheper:
@lafrech, 3.x is out now and it looks like they've addressed a lot of issues with the initial release. If you're otherwise swamped, I can take on cleaning up this pull request.

rbu:
> @lafrech, 3.x is out now and it looks like they've addressed a lot of issues with the initial release. If you're otherwise swamped, I can take on cleaning up this pull request.

That would be much appreciated. I have a code base that uses marshmallow 3 on pymongo directly and I'm evaluating if it makes sense to pull in umongo as an intermediary to get rid of some of our management code.

lafrech:
Indeed, we released MA3 in august and there's been a few fixes since then. Porting to MA3 would be great.

I won't do this right now, but you could start from where I stopped and I could try to find time to review before merging. It's a great changeset so two pairs of eyes won't be a luxury.

lafrech:
I just pushed a rebase of my MA3 port branch. This is still WIP.

@wfscheper, feel free to work on this if you get the time.

To get the tests to pass, I had two fixes in by marshmallow code:

- https://github.com/marshmallow-code/marshmallow/pull/1448 (not merged yet)
- I commented a few lines in Nested._serialize and _deserialize to revert and optim there that breaks our use case. More on this in https://github.com/marshmallow-code/marshmallow/issues/1424.

lafrech:
I pushed a new commit with a patch that makes all tests pass.

I'd rather solve this upstream, though. See https://github.com/marshmallow-code/marshmallow/issues/1424.

lafrech:
Rebased.

We could add the `Pluck` field, which was added in marshmallow 3, but this can be done later in a v3.x.

I'm tempted to release a 3.0.0b1 version and see how it goes.

@wfscheper @Cayke would you like to review / try that branch on your codebase?

wfscheper:
I'll try to look at this tonight.

Cayke:
> Rebased.
> 
> We could add the `Pluck` field, which was added in marshmallow 3, but this can be done later in a v3.x.
> 
> I'm tempted to release a 3.0.0b1 version and see how it goes.
> 
> @wfscheper @Cayke would you like to review / try that branch on your codebase?

Tried in my project and everything looks fine. I only had to refactor my fields because now you cannot set required and default params together.

lafrech:
`as_marshmallow_field` (more precisely `_extract_marshmallow_field_params`) only picks common field parameters.

We should also pick field-specific parameters. Each field can declare an attribute with a list of fields to pick.

I suppose this also affects umong2/marshmallow2 but I never caught got by it.

lafrech:
We should also check that parameters are correctly picked.

See https://github.com/marshmallow-code/marshmallow/issues/1281 and what happens in `_normalize_nested_options`.

It may not be safe to just get the attribute if it was changed in at init.

lafrech:
Or we could extend the `marshmallow_` mechanism to catch all `marshmallow_`-prefixed kwargs, not just a fixed list of expected attributes.

lafrech:
This is illustrated in https://github.com/Scille/umongo/commit/321d951ba3baa088f22db8ad7c23d0f7d2d3647f.

I'm tempted to change the prefix to just `ma_` for concision.

Note to self: Is the "Deserialize default/missing values" phase still needed with MA3?

lafrech:
I removed missing/default deserializaton because in MA3 those come in deserialized form.

This makes it more awkward to pass Dict or List as default, but since passing mutable default is a bad idea, I don't think it is that bad.

lafrech:
@wfscheper, @Cayke, @mymedia2, @Kilo59, I merged this.

There is still a few things to do before shipping umongo 3.

Before releasing a 3.0.0, I think we should release beta versions.

Meanwhile, you can already test master on your codebase and provide feedback.

Thanks.

lafrech:
I released a beta version.

Please try it when you get the time.

Also see https://github.com/Scille/umongo/milestone/1.

PR #157, opened 2019-01-10
Fix __getattr__ to deepcopy Document and EmbeddedDocument

lafrech:
I think the root cause is that when using `deepcopy` (I think this applies to pickle as well but didn't double-check), `hasattr(my_doc, __setstate__)` is called before the object is instantiated and therefore `self._data` does not exist, so we get an infinite recursion in `__getattr__`.

I think it is advised to add a condition like the one in this PR to fix this (inspired by https://stackoverflow.com/a/42272450/4653485).

Note that this prevents the use of attributes with leading and trailing double underscore, but oh well... I'm not sure this should even be considered a breaking change.

(CI errors unrelated and fixed in https://github.com/Scille/umongo/pull/155.)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ with open('HISTORY.rst', 'rb') as history_file:
     history = history_file.read().decode('utf8')
 
 requirements = [
-    "marshmallow>=2.6.0",
+    "marshmallow>=2.18.0",
     "python-dateutil>=2.5.0",
     "pymongo>=3.7.0",
 ]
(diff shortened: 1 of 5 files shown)


lafrech:
I don't think it hurts to merge this, but there is another issue with [deep|]copying: `missing` is duplicated.

This is a marshmallow issue. I proposed a fix in https://github.com/marshmallow-code/marshmallow/pull/1099. Hopefully, this will be fixed in marshmallow 3.0.

Until then, objects should not be copied/deepcopied.

Issue #83, opened 2016-12-19
Problem when modifying a Document pk


lafrech:
If I create and commit a Document, then change its id, I get an error when committing:

```python
    def test_create(self, classroom_model):
        Student = classroom_model.Student
        john = Student(name='John Doe', birthday=datetime(1995, 12, 12))
        ret = john.commit()

        print(john.id)
        # 58580041f667370f03a677e1
        print(john.is_created)
        # True
        john.id = ObjectId("5672d47b1d41c88dcd37ef05")
        # No error
        print(john.is_created)
        # True
        john.commit()
        # umongo.exceptions.UpdateError: <pymongo.results.UpdateResult object at 0x7f19b14d4a20>
```

This is because in `commit()`, we fall in the _update_ case and pymongo searches for a doc with the new id.

umongo should either prevent changing the ID field after the Document has been committed, or take necessary precautions if allowing it.

If we want to allow it, then maybe the role of `is_created` needs to be redefined, I don't know. At least, there should be a way to remember the old pk value so that the update can find the document.

I'm not sure this is a critical use case, but if feasible, I'd rather allow it.

Labelling as bug as it leads to an UpdateError exception while nothing tells the user he's doing something wrong.

touilleMan:
I would say modifying the `pk` field (i.e. the `_id` field) should be forbidden given then the `commit` operation would need to create a new document and delete the old one, which is something that can't be done in a single operation (see http://stackoverflow.com/questions/4012855/how-update-the-id-of-one-mongodb-document).

So we should throw an exception if user try such a thing, giving him a tip on what is the "right way" to perform this.

So far, I not really sure about this "right way" ;-)
I'm thinking of providing a `clone` method that would copy everything except the `pk` field. We could also tell the user to use this method when he tries to use copy/deepcopy on the Document object.

What do you think ?

lafrech:
I didn't know about this MongoBD "limitation". Throwing an exception makes sense, then.

As to the `clone()` method, yes, it could be useful.

Let's say that `clone()` **deep**copies everything but ID. Practically, this means that in the general case, a new ID will be created on commit, and if there is some ID field with default/missing attribute (most likely a callable, as a constant would not make sense), then ID is set to its default.

About the `__copy__` / `__deepcopy__`, implementing them is nice, unless they have an unexpected behaviour. Should they copy everything and let the user change the ID manually? Or should they behave like `clone()`?

[...]

PR #158, opened 2019-01-10
Add clone method to Document

lafrech:
Implements `clone` method proposed in https://github.com/Scille/umongo/issues/83#issuecomment-268842201.

This PR is based on https://github.com/Scille/umongo/pull/157 which fixes DataProxy deepcopy.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ with open('HISTORY.rst', 'rb') as history_file:
     history = history_file.read().decode('utf8')
 
 requirements = [
-    "marshmallow>=2.6.0",
+    "marshmallow>=2.18.0",
     "python-dateutil>=2.5.0",
     "pymongo>=3.7.0",
 ]
--- a/tests/test_document.py
+++ b/tests/test_document.py
@@ -377,6 +377,46 @@ class TestDocument(BaseTest):
         assert jane.child == john.child
         assert jane.child is not john.child
 
+    def test_clone(self):
+
+        @self.instance.register
+        class Child(EmbeddedDocument):
+            name = fields.StrField()
+
+        @self.instance.register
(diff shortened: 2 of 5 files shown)


lafrech:
If the document has an id field with a default value (typically a callable), perhaps the clone should get that default value.

This is tricky because it means finding and accessing that field instance.

Or taking a totally different approach using the schema to load the data. This is what I tried first, passing data proxy content as kwargs when creating the new document instance, but this raised other issues (dump_only fields,...) so I preferred to just copy the data proxy.

I don't see a trivial way to do that. I'll try to have a look at it. As a workaround, the user can override `clone` to add a default id...

lafrech:
I just pushed a new commit to solve the "default ID" case. I'm happy with this fix.

I don't really like playing with `DataProxy`'s internal `dict` from `Document` but I think it makes more sense to do this here rather than in DataProxy' itself, as the logic definitely belongs to `Document` and is not shared with `EmbeddedDocument`.

The line that resets `_modified_data` is probably useless because since the clone is a newly instantiated document, `is_created` is `False` and therefore, the next commit won't be an update.

lafrech:
See [comment ](https://github.com/Scille/umongo/pull/157#issuecomment-453531646) about `missing` singleton being duplicated when copying.

lafrech:
@touilleMan, the copy/deepcopy issue was fixed in marshmallow 2 branch, so we don't have to wait for umongo2 / marshmallow 3 to get this feature.

I'd like release umongo 1.1.0 with this. Any objection? (I'll update HISTORY.rst and bumpversion.)

touilleMan:
@lafrech go for it ;-)

lafrech:
Done!

Issue #161, opened 2019-01-19
Decimal128 supporting


serjshevchenko:
Could not create model with fields.Decimal.
When try to save document into mongo, error is raising
`umongo bson.errors.InvalidDocument: Cannot encode object: Decimal`
Mongo (pymongo) does not support Decimal data type. Since mongo 3.4, Decimal 128 data type is available.

lafrech:
Achieved in #162.

PR #162, opened 2019-01-19
Add Decimal128 support

serjshevchenko:
Fix Issue #161

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -615,5 +615,10 @@ class TestFields(BaseTest):
         assert d.to_mongo(update=True) == {
             '$set': {'in_mongo_price': Decimal128('8.1234')}}
 
+        d.from_mongo({'in_mongo_price': Decimal128('7.1234')})
+        assert d._data == {
+            'in_mongo_price': Decimal('7.1234')
+        }
+
         with pytest.raises(ValidationError):
             d.set('price', 'str')
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -583,6 +583,7 @@ class TestFields(BaseTest):
         d.load({'price': Decimal128('12.5678')})
         assert d.dump() == {'price': Decimal('12.5678')}
         assert d.to_mongo() == {'in_mongo_price': Decimal128('12.5678')}
+        assert d.get('price') == Decimal('12.5678')
 
         d.load({'price': Decimal('12.5678')})
         assert d.dump() == {'price': Decimal('12.5678')}
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -1,4 +1,6 @@
-from bson import ObjectId, DBRef
+from decimal import Decimal
+
+from bson import ObjectId, DBRef, Decimal128
 import pytest
 from datetime import datetime
 from dateutil.tz.tz import tzutc, tzoffset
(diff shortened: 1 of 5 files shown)


serjshevchenko:
As mentioned in Issue #161  descr., this only works since mongo3.4 cause only then Dicimal128 added.

umongo/fields.py
```diff
@@ -152,7 +152,12 @@ class IntegerField(BaseField, ma_fields.Integer):
 
 
 class DecimalField(BaseField, ma_fields.Decimal):
-    pass
```
lafrech: Please add a docstring with at least a warning about MongoDB version support (3.4+).

umongo/fields.py
```diff
@@ -152,7 +152,12 @@ class IntegerField(BaseField, ma_fields.Integer):
...
+    def _serialize_to_mongo(self, obj):
+        return Decimal128(obj)
+
+    def _deserialize_from_mongo(self, value):
+        if value:
+            return value.to_decimal()
```
lafrech: There should be an explicit else.

And what if the value is decimal `0`?

lafrech: Or perhaps it should just read

```py
    def _deserialize_from_mongo(self, value):
        return value.to_decimal()
```

tests/test_fields.py
```diff
@@ -570,3 +572,53 @@ class MyDoc(Document):
...
+
+        class MySchema(Schema):
+            price = fields.DecimalField(attribute='in_mongo_price')
+
+        MyDataProxy = data_proxy_factory('My', MySchema())
+        d = MyDataProxy()
+        d.load({'price': Decimal128('12.5678')})
+        assert d.dump() == {'price': Decimal('12.5678')}
+        assert d.to_mongo() == {'in_mongo_price': Decimal128('12.5678')}
+
```
lafrech: Missing

        assert d.get('price') == Decimal('12.5678')

?

lafrech:
I added the mashmallow<3 constraint in another commit for clarity.

@serjshevchenko, please

- rebase (remove the setup.py modification from your commit)
- add yourself to AUTHORS.rst
- add missing line in the tests (see my comment above)

And I think this is good to go.

AFAIU, this is not a breaking change as the Decimal was already broken on MongoDB < 3.4. I'm afraid the error for people using MongoDB<3.4 won't be very helpful. When trying to record a Decimal128 into a MongoDB 3.2 instance, I get:

    raise AutoReconnect("connection closed")
    pymongo.errors.AutoReconnect: connection closed

I have no idea how to do better.

@touilleMan, thoughts ? Any objection to releasing this as 1.2.0?

serjshevchenko:
after rebasing, coverage decreased.
Code which I added, is coverage fully.

lafrech:
Don't mind the coverage too much. It sometimes returns weird results. Covering your own change is enough.

@touilleMan, do you want to take a look before we push a 1.2.0 with this?

lafrech:
Alright. What's done can be undone. I took it upon myself to merge this and https://github.com/Scille/umongo/pull/165.

@touilleMan, I plan to release 1.2.0 by the end of the week.

touilleMan:
@lafrech sorry for the lack of reply (I spend a month in the Pilippines with barely no internet)
I have no objection on releasing 1.2.0. More generally, you're the most involved in this project currently and I trust your judgement so yo should feel free to release whenever you want ;-)

PR #173, opened 2019-03-07
Raise AlreadyCreatedError when modifying pk on a created document

### LINKED ISSUE #83: Problem when modifying a Document pk

lafrech:
Closes #83.

This change ensures an Exception is raised when modifying a document id after it is created.

I'm not really happy with this implementation. I don't think the initial issue is worth this change.

Feedback welcome. Better implementation even more welcome...

Also, the error message could be improved to suggest using `clone` (see #83).

### DIFF
--- a/tests/test_document.py
+++ b/tests/test_document.py
@@ -417,6 +417,19 @@ class TestDocument(BaseTest):
         assert jane.id != john.id
         assert jane.name == 'John Doe'
 
+    def test_modify_pk_field(self):
+
+        @self.instance.register
+        class User(Document):
+            primary_key = fields.ObjectIdField(attribute='_id', default=ObjectId)
+            name = fields.StrField()
+
+        john = User(name='John Doc')
+        john.primary_key = ObjectId()
+        john.is_created = True
+        with pytest.raises(exceptions.AlreadyCreatedError):
+            john.primary_key = ObjectId()
+
 
 class TestConfig(BaseTest):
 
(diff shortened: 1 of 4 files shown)


lafrech:
Idea for a better implementation: add a method/property to the data proxy returning id field name, and make the check/raise in document.

lafrech:
Superseded by #258.

PR #179, opened 2019-03-14
Only use async/await coroutine style in asyncio framework

touilleMan:


lafrech:
Thanks for tackling this.

Looks like you're using asynchronous comprehensions, introduced in Python 3.6:
https://docs.python.org/3/whatsnew/3.6.html#pep-530-asynchronous-comprehensions

lafrech:
I fixed the 3.5 compatibility issue. Tests pass.

I can't really provide an educated review as I'm async illiterate. I'll trust you on this one and merge as is.

touilleMan:
@lafrech hooop I didn't notice your fix !

I though we were also dropping Python 3.5 so I pushed --force and destroyed your fix ! (I should have used `--force-with-lease`...)

Can you re-push this branch with your fix please ?

lafrech:
Ah ah. No pb. Done.

It takes a little more to run tests on 3.7. We need to tell Travis to use Xenial. I'll do that if you want when this is merged. https://github.com/Scille/umongo/issues/180

PR #181, opened 2019-03-14
Support Python 3.7

### LINKED ISSUE #180: Support Python 3.7

lafrech:
Closes #180.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
-sudo: false
+language: python
+dist: xenial
+cache: pip
+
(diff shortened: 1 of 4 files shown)


lafrech:
We could drop the 3.6 tests to limit the load and testing time. If most and least recent are covered, we're fine.

It used to be only 3.4 and 3.6 before we dropped 3.4 and I replaced it with 3.5.

touilleMan:
@lafrech agree with only testing 3.5 and 3.7

PR #186, opened 2019-03-20
Don't call _deserialize_from_mongo in Field._deserialize

lafrech:
For the reason this was done in the first place, see https://github.com/Scille/umongo/pull/84#issuecomment-268840412.

I just stumbled on a use case where it wouldn't work.

The assumption that deserialization (from outside to OO world) provides what pymongo would return is incorrect when the data is not serialized in DB in its OO world form. This is the case for fields that have custom `_serialize_to_mongo` / `_deserialize_from_mongo` methods.

For instance, if the embedded document contains a `DateField`, the deserialization produces a `datetime.date` and `deserialize_from_mongo` will choke on that because it expects a `datetime.datetime`.

I believe this fix does the right thing.

The only thing that bothers me is that it duplicates the `_schema` string from marshmallow:

```py
    ValidationError({'_schema': ['Invalid input type.']})
```

but this is public API, so it's not a big deal. It won't change in a minor version.

### DIFF
--- a/tests/test_embedded_document.py
+++ b/tests/test_embedded_document.py
@@ -212,8 +212,9 @@ class TestEmbeddedDocument(BaseTest):
         assert child.to_mongo() == {'in_mongo_a_child': 1, 'b': 2, 'c': 3, '_cls': 'EmbeddedChild'}
         assert grandchild.to_mongo() == {'d': 4, '_cls': 'GrandChild'}
 
-        with pytest.raises(ValidationError):
+        with pytest.raises(ValidationError) as exc:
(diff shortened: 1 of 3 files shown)


lafrech:
I went ahead and refactored the other fields doing the same thing.

Since `EmbeddedDocument._deserialize` does not call `_deserialize_from_mongo` anymore, I could revert the trick I introduced in https://github.com/Scille/umongo/pull/110.

Should be cleaner now.

@touilleMan, I'd like to release this as 2.0.1 soon.

PR #187, opened 2019-03-25
Added link to documentation

borenstein:
Took me awhile to find documentation because it wasn't linked on the GitHub README. I kept looking because I could see that the program was really well documented. Once I found it, I thought others might be having the same problem, so I propose to add the link.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,6 +16,8 @@
     :target: https://coveralls.io/github/Scille/umongo?branch=master
     :alt: Code coverage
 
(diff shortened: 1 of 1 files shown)


touilleMan:
Thanks for the contribution !

Could you modify your PR in order for your hyperlink to appear as a nice text (something like "see documentation & tutorial") ?

borenstein:
Sorry, I don't have time, but I'm glad it's on your radar now

PR #195, opened 2019-05-15
Fix is_modified

lafrech:
Various fixes about `is_modified`:

- Include modified `BaseDataObject`s in `BaseDataProxy.get_modified_fields` and `BaseDataProxy.get_modified_fields_by_mongo_name`. While fixing this, I factorized the logic in `get_modified_fields`. I don't think it is much worse in terms of performance. BTW, there was a type mismatch in those two methods, as the first returned a `list` and the second returned a `set`. Now, both return a `set`. I don't think this qualifies as a breaking change since it is not public API. `BaseDataProxy.get_modified_fields_by_mongo_name` was not even tested.

- Always return a boolean in `List.is_modified`. Previously, it could return `None`.

- `List`: set modified when using `del my_list_field [i]` (Issue reported in https://github.com/Scille/umongo/issues/71#issuecomment-489995372).


lafrech:
Tests pass.

@touilleMan I intend to merge/release as 2.0.4 soon.

PR #206, opened 2019-10-02
add __setitem__ to dict data object

otistamp:


lafrech:
setitem is not the only operation. If we go this way, we'll want to also override delitem, pop, etc.

Also, the code here is wrong because it does not call `super().__setitem__`.

And I wouldn't test for equality, if only for consistency with what is done in `List`.

Overall, unless we have a complete change set with the other `dict` methods supported, I'd rather leave it as it is. Less confusing to assume Dict does not do anything than having it do half the job.

Things should go smoother when umongo moves to marshmallow 3 as `Dict` accepts inner fields. I don't have time right now for this, so if anyone wants to pick up the porting work, please let me know.

otistamp:
ok, yeah. gonna close this in favor of porting to marshmallow 3.

Issue #163, opened 2019-01-23
I tried implementing Motor Tornado based on Motor Asynio and it works, is this correct implementation ??


OneWithNoFace:
[motor_tornado.py.txt](https://github.com/Scille/umongo/files/2786210/motor_tornado.py.txt)

Changes made in umongo/frameworks/motor_asynio.py:
1) Imported corresponding class (e.g, AsyncIOMotorDatabase -> MotorDatabase)
2) Changed class names accordingly.

Is this way correct ??

touilleMan:
Hi,

>  So I thought that changing the import according to the motor tornado would work. I tried by creating a simple tornado application and it does work for the creating a Document instance and committing it.

Yes, that's my idea also. Beside if you have also converted the framework tests and everything is passing I guess you have something worth merging ;-)

Can you open a PR with this code please ?

zikphil:
I have been running the AsyncIO Instance with Tornado for a couple months now on Production and it works fine. Am I missing something?

lafrech:
I have no time (and interest) for this right now.

Did you check the diff between AsyncIO (in the code) and Tornado (in your PR)?

touilleMan:
@zikphil If I recall well, Tornado is now compatible with the asyncio event loop.
This used to be not the case (for instance the motor driver provides 2 different backends for asyncio and for Tornardo)
So it's very possible Tornado is now supported out of the box by just supporting asyncio \o/ (if that's the case it would be good to update the document and the README)

I guess it would be still interesting to add a new `test_motor_tornado.py` which would be heavily based on https://github.com/Scille/umongo/blob/master/tests/frameworks/test_motor_asyncio.py (basically just start a tornado event loop instead of an asyncio one) to make sure there is no cornercases

lafrech:
I'm happy to merge a PR doing this.

PR #216, opened 2020-01-07
Adding support for Tornado motor async client

zikphil:
I just don't want this to get lost so I am putting it here, I've implemented the solution described in #163 to support motor's Tornado Async Client.

Sorry for the poor structure or lack of testing, just wanted to make sure this was visible to other people wanting to implement this.


lafrech:
I'll defer to @wfscheper as I don't work with async code.

We'd be happy to merge a PR if you or anyone is willing to submit code + tests.

PR #220, opened 2020-02-03
Added None check to avoid nested NoneType exception

kunla:
I was able to recreate this error with pytest by attempting to insert a record into mongo with a pre-existing ObjectId.

### DIFF
--- a/umongo/frameworks/pymongo.py
+++ b/umongo/frameworks/pymongo.py
@@ -121,6 +121,9 @@ class PyMongoDocument(DocumentImplementation):
                 self.is_created = True
                 self.post_insert(ret)
         except DuplicateKeyError as exc:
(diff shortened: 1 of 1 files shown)


lafrech:
Thanks.

Can you please elaborate? IIUC, the `DuplicateKeyError` is supposed to have a `details` but it doesn't. How come?

Also, I guess the fix should be applied to all frameworks (e.g. https://github.com/Scille/umongo/blob/master/umongo/frameworks/txmongo.py#L88).

kunla:
I took a closer look and it appears that this only occurs in MongoMock - the other frameworks are working correctly.

lafrech:
Closing as unclear to me. Please comment if you're still interested in working on this.

PR #221, opened 2020-02-12
feat: umongo, sanic and motor integration example.

humbss:
Adding Sanic example.

Problem: A fully integrated example of umongo + sanic.

Solution: This PR contains the example using Motor as driver, the example contains a layered application from rest API to model persistence thru uMongo ODM.


lafrech:
Thanks for sharing this.

I'm reluctant to merge this as is.

As you can see in other examples, we try to keep examples as small minimal as possible.

This one involves unrelated stuff with opinionated choices, such as deploying mongo with docker, which IMHO is out of scope.

It is nice to provide a full working example, but the more you expand the scope to external dependencies, the sooner the example is outdated due to some change there.

This is totally worth a gist, though.

And if you're willing to work on this to share it here, perhaps you could focus on umongo + sanic and build a smaller API so that it fits in one or two files.

humbss:
Hi, I agree with you, I'll submit a minimal example to handle this.

thx

Issue #185, opened 2019-03-20
Building a query from OO world values


lafrech:
umongo tries to be close to pymongo when building queries, to avoid complicated logic. Pretty much all it does is map OO world names to Mongo world names (using `field.attribute` ) and manage inheritance by adding `_cls` if needed.

I recently found out that there are other limitations we may want to address.

When building the query, the user will generally enter data in OO world form:

```py
    date = dt.date(2014, 5, 12)
    number = 12
    Document.find_one({'birthdate': date, 'quantity': number})
```

This works most of the time but it will fail when the Mongo world form differs from the OO world form. That is, when the field has a custom `_serialize_to_mongo` method.

Such fields are:

- `DateField` and `DecimalField`. Those fields are recent additions. Since there is no native BSON format for them, they are cast in a specific format when serializing to mongo (resp. `datetime.datetime` and `bson.Decimal128`. Passing them in OO world form in the query results in a pymongo error.

- Container fields such as `DictField`, `ListField`, `GenericReferenceField`, `EmbeddedField`. For those fields, an error will be raised if the query is entered in object form, but this may not be a big deal as the user "naturally" enters the query in Mongo world form (as a `list` or `dict`):

```py
    Document.find_one({'embedded': {'field': value}})
```

- `ReferenceField`. Likewise, an error is raised when passing a `Document` instance rather than an id (see https://github.com/Scille/umongo/issues/169).

----------------------------------------------------------------------------------------------------------------------

An approach to address this would be to get to the field and use it to serialize the value before passing the query. This is what I've been trying to achieve.

```py
def is_operator(entry_or_query):
    return isinstance(entry_or_query, str) and entry_or_query.startswith('$')

def map_entry_with_dots(entry, entry_query, fields):
    """
    Consider the given entry can be a '.' separated combination of single entries.
    """
    print(list(fields))
    mapped = []
    for sub_entry in entry.split('.'):
        field = fields.get(sub_entry)
        mapped_sub_entry, entry_query, fields = map_entry(sub_entry, entry_query, fields)
        mapped.append(mapped_sub_entry)
    # Serialize value to Mongo world
    if field is not None and not is_operator(entry_query):
        entry_query = field.serialize_to_mongo(entry_query)
    return '.'.join(mapped), entry_query, fields

def map_query(query, fields):
    """
    Retrieve given fields whithin the query and replace there name with
    the one they should have within the database.
    """
    if isinstance(query, dict):
        mapped_query = {}
[...]

PR #245, opened 2020-04-16
Use fields for keys/values in DictField

### LINKED ISSUE #99: DictField limitations
`DictField` suffers from a few limitations.

- Must be marked as modified manually
- Does not deserialize values into objects (no Schema)
- Can't use non-string keys

It might be useful to have a schema-less structure, but sometimes, I'd like to have `DictField` enforce a schema. Maybe we could add it extra features or create a `TypedDictField` (or `StrcturedDictField`, I don't know about the naming).

See [this discussion](https://github.com/marshmallow-code/marshmallow/issues/483) in Marshmallow about nesting a `Schema` into a `Dict`.

[This comment](https://github.com/marshmallow-code/marshmallow/issues/483#issuecomment-285936314) illustrates the limits of `DictField`.

The enhancement proposal is to add the possibility to do this:

```python
    fields.Dict(keys=fields.Str(), values=fields.Nested(SomeSchema))
```

Benefits:

- Serialization/deserialization of keys and values
- Allows to use non-string keys (hashable and serializable)
- Keep track of is_modified, no need to mark the field as modified manually

Note: uMongo does not complain if a Dict is passed non-string key, but pymongo will raise an `Exception` at commit time. This could be protected by a validator at assignment time in current `DictField`. Obviously, adding the feature described above would solve this by allowing non-string keys.

Not sure which approach would be best. Try to solve upstream https://github.com/marshmallow-code/marshmallow/issues/483 and ultimately if rejected implement `TypedDictField` as a marshmallow-bonus field, or create `TypedDictField` here as a proof-of-concept and try to have it integrated upstream. The more we can provide upstream, the best, and I saw no objection to the idea, there.

lafrech:
Closes #99.

### DIFF
--- a/tests/test_data_proxy.py
+++ b/tests/test_data_proxy.py
@@ -462,16 +462,23 @@ class TestDataProxy(BaseTest):
 
         class MySchema(EmbeddedSchema):
             # EmbeddedField need instance to retrieve implementation
-            listed = fields.ListField(fields.EmbeddedField(MyEmbedded, instance=self.instance))
-            embedded = fields.EmbeddedField(MyEmbedded, instance=self.instance)
             required = fields.IntField(required=True)
+            embedded = fields.EmbeddedField(MyEmbedded, instance=self.instance)
+            listed = fields.ListField(fields.EmbeddedField(MyEmbedded, instance=self.instance))
+            dicted = fields.DictField(
+                values=fields.EmbeddedField(MyEmbedded, instance=self.instance))
 
         MyDataProxy = data_proxy_factory('My', MySchema())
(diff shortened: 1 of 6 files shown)


lafrech:
This implementation allows the user to optionnaly use fields for keys and/or values.

Any type of field can be used, so it is up to the user not to pick a field that doesn't make sense for a key. Generally, fields that serialize immutable stuff serialize as string, so it should work, but then the user should be careful when querying (see https://github.com/Scille/umongo/issues/99#issuecomment-286040344). Nothing really new (see https://github.com/Scille/umongo/issues/185).

PR #259, opened 2020-04-26
Use txmongo's find_with_cursor method

lafrech:
This PR fixes a deprecation warning due to [API changes](https://txmongo.readthedocs.io/en/latest/NEWS.html#api-changes) in txmongo.

I had to bump minimal txmongo version to 19.2.0 (latest) for the tests to pass. I think we should be able to support older versions. I suspect I just didn't find the proper pymongo, twmongo and perhaps twisted and pytest-twisted combination. The incomplete changelog didn't help. I went through the commits in txmongo and couldn't find the reason versions prior to 19.2.0 would fail.

I don't mind bumping this since it is only a test dependency, and we're releasing a major version, but for the sake of understanding and because it would be nice to know which txmongo versions are supported, explanation welcome.

BTW, @wfscheper, I hope you're doing well and I'd be glad to get some feedback on umongo 3 beta if you can find the time to give it a try.


lafrech:
I forgot to bump txmongo version in setup.py.
Should be at least 16.3.0, maybe more (see above).

lafrech:
Let's move on with this.

PR #261, opened 2020-04-28
Add ExposeMissing context manager

lafrech:
- Use Python 3.7's context variable feature to create a context manager in which a document returns missing for missing data rather than None
- Use this context manager in schema_from_umongo_get_attribute

This makes the Document responsible for outputing missing rather than the schema having to guess. Also the schema works with any type of data, not only a Document instance (with other data, the context manger is no-op).

I hoped to remove the whole `mongo_world` management thanks to this, but there are other issues with data from Mongo vs. Document.

The benefit is not as great as I expected, but I still like this implementation.

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -569,7 +569,7 @@ from scratch... almost:
     >>> MassiveDuckSchema(marshmallow.Schema):
     ...     breeds = marshmallow.fields.List(marshmallow.fields.Nested(MassiveBreedSchema))
 
-.. note:: A custom marshmallow schema :class:`umongo.marshmallow_bonus.SchemaFromUmongo`
+.. note:: A custom marshmallow schema :class:`umongo.schema.RemoveMissingSchema`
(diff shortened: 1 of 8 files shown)


lafrech:
This implementation is totally sub-optimal. It misses a nice point of the context variable, which is that we could put the context manager at the dump method level rather than the get_attribute level. This is another reason to like the context_var approach better than the current one.

lafrech:
Rebased on master.

- Now uses the context manager in schema dump method.
- Removes now useless `schema_from_umongo_get_attribute`.
- Exposes (`SchemaFomUmongo` renamed as) `RemoveMissingSchema`.
- Makes `RemoveMissingSchema` an opt-in. Easy to achieve by passing it to custom `BaseDocument` and `BaseEmbeddedDocument`.

lafrech:
Finally, this is the first part of a rework followed in #299 that removes `mongo_world` and nicely simplifies the code.

Issue #271, opened 2020-05-07
Mixin feature doesn't support attribute access


lafrech:
The mixin support introduced in #267 (umongo 3.0.0b6) is broken: it does not support attribute access as `__getattr__` is not entered since the field is accessed first.

I'd rather not play with `__gettattribute__`.

I guess the feature is more complicated than I expected. The mixin class attribute are correctly fetched in `_collect_schema_args` but the mixin still appears in the MRO so the normal attributes are duplicated (from the mixin to the class) and the fields are shadowed by the mixin fields.

We could modify `_convert_bases` to remove the mixin class (in fact, every non-implementation class) from the MRO but this seems to go against the zen to get everything from the class then drop it. For instance, `isinstance(my_class, Mixin)` will return `False` which seems wrong. This is accepted for documents because it is explicit and because the user will check against the implementation class, not the template class.

I'm afraid the cleanest way to allow mixin classes to provide fields would be to add a third class of document, `MixinDocument`, with template and implementation, registration process, etc... Perhaps `MixinDocument` could be a base for `EmbeddedDocument`, which itself could be a base for `Document` (see #266).

For now, I'm tempted to just revert the feature and see how the Doc/EmbeddedDoc factorization goes.

PR #272, opened 2020-05-07
Replace (Embedded)Document.real_attributes with _fields

lafrech:
`_fields` is computed once on class creation rather than on first access
Allows setting arbitrary attributes on objects

----------------------

My first intent here was to fix an issue with fields from mixin classes. It appears the problem is more complicated than this (#271) so I'm not sure this change is worth it.

I don't really see the point of forbidding setting arbitrary attributes, so I'm still fine with this change. But I don't mind keeping things as they are as well.

I like the idea of setting the list at class creation time rather than on access, but I guess it could be achieved with real_attributes as well to keep current behaviour.

Note: See if to_raise argument in data proxy methods still has a meaning.

### DIFF
--- a/tests/test_document.py
+++ b/tests/test_document.py
@@ -102,11 +102,14 @@ class TestDocument(BaseTest):
         with pytest.raises(AttributeError):
             john.missing
         with pytest.raises(AttributeError):
-            john.missing = None
-        with pytest.raises(AttributeError):
             del john.missing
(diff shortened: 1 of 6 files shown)


lafrech:
Removed `to_raise`. I like the new logic. And I think setting arbitrary arguments is fine. Let's move on with this.

PR #288, opened 2020-07-13
Fixed `allow_none` on embedded list

fsonntag:
I realized, that when I have a `fields.ListField(fields.EmbeddedField(MyEmbedded), allow_none=True)`, a `TypeError` is raised, when `enumerate(value)` in `_required_validate(...)` is called. 
So I added another check for the `allow_none` attribute and skip this per-item validation.

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -88,6 +88,19 @@ class TestRequired(BaseTest):
         doc.required_validate()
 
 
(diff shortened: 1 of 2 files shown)


lafrech:
Thanks.

I reworked this in #289.

fsonntag:
Thanks a lot :) I will close this PR then.

PR #296, opened 2020-09-03
Fix pylint6 compat (skipif requires a string)

lafrech:


atti92:
you were faster.

lafrech:
Done in 2.x branch then merged in master (without a PR).

PR #309, opened 2020-11-02
Use structured error info from DuplicateKeyError

lafrech:
Use structured info rather than parse error message.

No need to iterate over indexes. (This change could be a step towards solving #79.)


lafrech:
Just noticed the failing tests. There were Travis issues so I ignored the tests before merging as tests passed on my setup.

Seems to be MongoDB version dependent. I shall investigate.

lafrech:
This change uses information added in MongoDB 4.2 

https://github.com/mongodb/mongo/commit/410656e971aff8f491a87337a17d04bd866389ba
https://jira.mongodb.org/browse/SERVER-14801

So we're basically dropping MongoDB 4.0 and older.

Not really an issue but let's make it clear with #313.

PR #311, opened 2020-11-05
Use Azure DevOps for the tests and support Python 3.9

lafrech:
The coverage is wrong because results from separate builds are not merged (https://github.com/asottile/azure-pipeline-templates/issues/36).

Tests are now run on MongoDB 4.4 (Travis runs MongoDB 4.0). Could be worth investigating testing on different versions.


lafrech:
Let's remove the coverage badge for now as the coverage report is inaccurate.

PR #314, opened 2020-11-10
Rework instance

lafrech:
Remove `templates` argument to `BaseInstance` (undocumented and untested).
Merge `BaseInstance`, `Instance` and `LazyLoaderInstance` into `Instance`.
Make `Instance` an abstract base class.
Rename `Instance.init` into `Instance.set_db`.
Don't expose framework instances in `umongo` top module. They must be imported from `umongo.frameworks`.

Before this change, one could use `Instance(db)` and get the right builder class, but not the right instance class, which works fine because there is no specific method in each framework instance class (for now).

After this change, one must pick a specific framework instance or explicitly call the factory method `Instance.from_db`.

The use case of registering documents then setting the db didn't work before the change because one would either call `Instance(db)` or pick the dedicated framework instance. The only difference between the two is that the latter allows to pass the db object after registering documents. But still, the instance type must be defined.

I'm keeping the `Instance.from_db` factory method in case it is useful to anyone, but I don't really see the use case. Since the db object exists, the user probably knows what framework he is using, so he could pick the right framework instance himself.

Lazy loading still works: one can instantiate a framework specific instance, register documents, and then call `set_db` to actually pass a DB connection. Typical use case, the flask app factory: load all model code and register all documents in the instance at import time, then in app init, set a DB connection.

### DIFF
--- a/docs/apireference.rst
+++ b/docs/apireference.rst
@@ -9,11 +9,8 @@ API Reference
 Instance
 ========
 
-.. autoclass:: umongo.instance.BaseInstance
-    :inherited-members:
-
-.. autoclass:: umongo.Instance
-  :members:
+.. autoclass:: umongo.instance.Instance
+    :members:
 
 .. autoclass:: umongo.PyMongoInstance
 
--- a/docs/apireference.rst
+++ b/docs/apireference.rst
@@ -13,8 +13,6 @@ Instance
     :inherited-members:
 
 .. autoclass:: umongo.Instance
-
-.. autoclass:: umongo.instance.LazyLoaderInstance
   :members:
 
 .. autoclass:: umongo.PyMongoInstance
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,7 +1,7 @@
 import pymongo
 
 from umongo.document import DocumentImplementation
-from umongo.instance import Instance, LazyLoaderInstance
+from umongo.instance import Instance
 from umongo.builder import BaseBuilder
 from umongo.frameworks import register_instance
 
@@ -58,7 +58,7 @@ class MockedBuilder(BaseBuilder):
     BASE_DOCUMENT_CLS = DocumentImplementation
 
 
-class MockedInstance(LazyLoaderInstance):
+class MockedInstance(Instance):
     BUILDER_CLS = MockedBuilder
 
     @staticmethod
--- a/tests/common.py
+++ b/tests/common.py
@@ -3,7 +3,7 @@ import pymongo
 from umongo.document import DocumentImplementation
 from umongo.instance import Instance, LazyLoaderInstance
 from umongo.builder import BaseBuilder
-from umongo.frameworks import register_builder
+from umongo.frameworks import register_instance
 
 
 TEST_DB = 'umongo_test'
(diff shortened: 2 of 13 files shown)


armsp:
@lafrech Do you think what you have done here solves this issue in this [SO post](https://stackoverflow.com/questions/64879897/how-to-separate-umongos-object-document-models-from-the-main-web-app-or-how-to)?

I am not sure but I think I am trying to figure out how to separate instance from the document object model i.e. the instance is declared in a main.py file that also starts a web app and the model definitions stay in a separate file. Is it possible to have something like that?

lafrech:
See my answer on SO.

This PR does not change anything to it, apart from the `init` method being renamed to `set_db`.

armsp:
Hi @lafrech, I am a little stuck. I haven't been fully able to understand some differences hence I haven't had the chance to accept your answer. 
I still haven't been able to make my application work.
Based on the documentation, I had initially setup things like -
```py
client = MongoClient(host='web-db_1', port=27017)
database_name = os.environ.get("DATABASE")
db = client[database_name]

instance = Instance(db)
app = FastAPI()
```
Which didn't work ofcourse and gave me the following error -
```
web_1     | umongo.exceptions.NoDBDefinedError: init must be called to define a db
```

Now I didn't want to copy and paste the code without understanding. My current state of things is this -
```py
client = MongoClient(host='fastapi-tdd-docker_web-db_1', port=27017)
database_name = os.environ.get("DATABASE")
db = client[database_name]
instance = PyMongoInstance()
instance.init(db)
app = FastAPI()
```
But this fails at the code where I commit data to my db and basically gives the exact same error as above but at the location of commit code rather than in the main.py file.

In the solution that you provided, you ask me to do something like the following -
```
instance = PyMongoInstance()
database = MongoClient()
instance.init(database)
```
In the docs I often see the following (I am also more familiar with this coming from MongoEngine) -
```
from pymongo import MongoClient
from umongo import Instance
client = MongoClient()
db = client.test
instance = Instance(db)
```
Could you please tell me the difference b/w Instance, `MongoClient` and `PyMongoClient`, and which one should I actually use? Also, you seem to be using `instance` differently than mentioned in the docs.

lafrech:
Sorry, I meant 

```py
instance = PyMongoInstance()
database = MongoClient().test  # <- I forgot .test here. SO answer fixed.
instance.init(database)
```

`MongoClient` is the driver from pymongo, providing connection to MongoDB instance. `MongoClient.test` is a connection to the `test` DB, as in connect to MongoDB, then `use test`. An instance is a umongo concept allowing to use different database drivers. A `PyMongoInstance` is an `Instance` of pymongo type, with methods specific to pymongo when it comes to reading/writing in DB.

`Instance(db)` should work with umongo 2 and umongo 3 beta until 3.0.0b12. The reason for the change are explained above.

I don't know why your first code block above doesn't work.

I suspect, perhaps, your instantiating an `Instance` in this code block somewhere in your main application file, but you're registering your documents with another instance created elsewhere in the code.

You should of course set the DB on the same instance instance (!) than the one used to register documents. This is why in my SO answer, I'm showing you how to create the instance in a common file imported in both main app file and model files.

lafrech:
This PR introduces a bug fixed in #318.

PR #319, opened 2020-11-23
Add migration instance classes

lafrech:
umogo 3 introduces a change in the way embedded documents are represented in database.

This PR provides instance subclasses that can help a user migrating his DB.

The process is not really automatic, but at least this class is useful to build a simple migration script.

Note that it requires access to the model (umongo classes). In other words, the easiest migration path would be to branch the application code to plug this instance in place of the usual one, call the migration method, then end the application.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,6 +12,7 @@ Contents:
    :maxdepth: 2
 
    userguide
+   migration
    apireference
    contributing
    authors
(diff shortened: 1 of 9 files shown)


lafrech:
Reworked version only uses the model to determine which `EmbeddedDocument` classes names should be removed, then recurses on all documents to remove those keys using plain pymongo. It should have much less side effects.

Although it is idempotent, it does not make any sense to do it on each app startup and it should be used in a dedicated one-shot script.

PR #323, opened 2020-12-07
Make BaseSchema ordered

### LINKED ISSUE #55: Is there support for OrderedDict
If you like to find embedded documents the order of the kv pairs in the embedded document is essential as 'find' will only match documents with the same order. As the mongo docs say (https://docs.mongodb.com/manual/tutorial/query-documents/#exact-match-on-the-embedded-document):

> Equality matches on an embedded document require an exact match of the specified <value>, including the field order.

pymongo thus supports python OrderedDict: 
http://stackoverflow.com/a/30787769/4273834

as does marshmallow:
http://marshmallow.readthedocs.io/en/latest/quickstart.html#ordering-output

Is there a way to use OrderedDict in umongo?

lafrech:
Closes #55. Fixes broken embedded document query.

Might introduce a performance penalty.

We could make the base class customizable and this opt-in but
- it would add complexity
- the embedded doc query would still be broken by default

We could make the base class customizable and this opt-out but
- it would add complexity
- people with a real will against this (because of the perf penalty and because they don't query on embedded docs) can monkey-patch the base class.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -79,3 +79,8 @@ class BaseDBTest:
 
     def setup(self):
         con.drop_database(TEST_DB)
+
+
+def assert_equal_order(dict_a, dict_b):
+    assert dict_a == dict_b
+    assert list(dict_a.items()) == list(dict_b.items())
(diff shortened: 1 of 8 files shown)


lafrech:
I'm willing to move on with this. The performance impact should be minimal in most cases and I'd rather favor functionality over performance.

At this point, I'm tempted to also make generated pure marshmallow schemas ordered as well. For consistency. And because it is nice from an API perspective. I'm adding a commit to do so. I can be opt-out by replacing the base schema.

PR #331, opened 2021-01-04
Update credits and contact

lafrech:
Updating credits.

I don't mean to kick anyone out, just trying to be accurate, especially in the contact email, so guys please tell me if this suits you.

You can't really put two authors in setup.py. See https://stackoverflow.com/a/10005265/4653485 and links.


touilleMan (approved):
:+1: :tada:

Issue #116, opened 2017-08-24
Add how to set primary key in documentation


defrank:
I assumed there was no support for specifying a field as a primary key, but all one needs to do is set the `_id` field.  This should be made explicit in the documentation.

Coming from other ORM/ODMs, I figured something more like `email = EmailField(primary_key=True, required=True, unique=True)` was expected.

```python
class User(Document):
    _id = EmailField(required=True, unique=True)
    first_name = StrField()
    last_name = StrField()

    @property
    def email(self):
        return self._id

    @email.setter
    def email(self, value):
        self._id = value
```

lafrech:
See http://umongo.readthedocs.io/en/latest/userguide.html#mongo-world

I believe case 1 can be written.

```python
class User(Document):
    email = EmailField(attribute='_id')
    first_name = StrField()
    last_name = StrField()
```

Does this generate a ReferenceField validation issue? That would be a bug, I guess.

touilleMan:
As @lafrech said, this is already covered

> This almost works which currently works, but ran into validation issues when the object is used in a ReferenceField since _id is not a valid ObjectId.

This however is a nasty little bug ! Can you provide a stacktrace and a test case ?

defrank:
Ah, didn't notice the `attribute` argument.  Perhaps an explicit mention of overriding the primary key could be added in the docs?  Thanks for pointing it out.

Basically my use case:
```python
@instance.register
class Network(Document):
    _id = StrField(required=True, unique=True)

    @property
    def name(self):
        return self._id

    @name.setter
    def name(self, value):
        self._id = value

    class Meta:
        allow_inheritance = True


@instance.register
class NetworkContract(EmbeddedContract):
    network = ReferenceField(Network, required=True)

    class Meta:
        allow_inheritance = True


@instance.register
class Entity(Document):
    emails = ListField(EmailField())
    network_contracts = ListField(EmbeddedField(NetworkContract))

    class Meta:
        allow_inheritance = True


if __name__ ==  '__main__':
    network = Network(_id='facebook')
    network.commit()
    entity = Entity()
    entity.network_contracts.append({'network': network})
```

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/derek/Code/fanai-system/venv/lib/python3.6/site-packages/umongo/data_objects.py", line 25, in append
    obj = self.container_field.deserialize(obj)
[...]

PR #343, opened 2021-02-03
Allow Reference to be any kind instead of just ObjectId

jbkoh:
This PR addresses https://github.com/Scille/umongo/issues/116, with the expense of `Reference` not being a subclass of `ObjectId`. `Reference` should be able to be used for other field types such as `StringField`, so I'd argue `Reference` is not a subclass of `SubjectId`. 

pytest is successful except for the test to validate `Reference` being a subclass of `ObjectId`. I haven't not changed the test to discuss this further, but I'd be happy to remove that part if the maintainers agree.

This PR implements mutation of `Reference` type by introducing `_pk_template` variable (https://github.com/jbkoh/umongo/blob/fix-referencefield/umongo/fields.py#L332). It is lazily evaluated at https://github.com/jbkoh/umongo/blob/fix-referencefield/umongo/marshmallow_bonus.py#L45 by inspecting the `document_cls`'s `pk_field`. If `pk_field` is not present at the template, it means that it's a default type, `ObjectId`.

Again, this passes all the tests except `Reference` being `ObjectId` which can be removed, I think.

Please let me know if this is the right direction which we can discuss further. If it's completely out of track, I'll just abandon this one.

Thanks!

### DIFF
--- a/umongo/fields.py
+++ b/umongo/fields.py
@@ -329,6 +329,8 @@ class ReferenceField(BaseField, ma_bonus_fields.Reference):
         self._document_cls = None
         self._document_implementation_cls = DocumentImplementation
 
+        self._pk_template = None  # This will be lazily defined at serialization.
+
     @property
     def document_cls(self):
         """
--- a/umongo/marshmallow_bonus.py
+++ b/umongo/marshmallow_bonus.py
@@ -1,6 +1,9 @@
 """Pure marshmallow fields used in umongo"""
 import bson
 import marshmallow as ma
+from .template import get_template
+
+from bson import ObjectId as BsonObjectId
 
 from .i18n import gettext as _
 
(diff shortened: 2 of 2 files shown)


jbkoh:
I realized it doesn't properly implement the non-objectid references yet. I'm looking into it, but I'd be happy to receive any advice on fly.

lafrech:
Thanks for working on this.

I can't review right now.

I had begun work on this: https://github.com/Scille/umongo/tree/reference.

I think my approach is different:

- Remove Reference marshmallow field
- Modify umongo ReferenceField so that the marshmallow field used inside is the appropriate field accord to the pk type

The difficulties I faced:

- GenericReferenceField is more difficult because you don't know the type in advance
- Generated marshmallow schema must include the proper marshmallow field, wrapped in some reference layer. This must be generated dynamically, ideally with some sort of caching to avoid generating N times the same schema.
- The most difficult part is to generate the marshmallow schema for the GenericReferenceField because the type is unknown, so it is undefined. Maybe we can just pass a Raw field or by default an ObjectId field and document the limitation.

jbkoh:
Sorry I'm closing this one as the generic implementation will need more effort than what I could contribute. Still this is a great project and I'm actively using it. Thanks!

PR #354, opened 2021-06-30
Support Type Hints

### LINKED ISSUE #233: Support for type hints
Hi,

I'm thinking about adding something to provide static type support. To be able to use umongo with mypy for example.

I think there is not really an easy way to provide this kind of feature since each DocumentImplementation method can have it's own signature.
Therefore, each implementation will be type annotated. For now, I'm unsure this will be enough for a mypy commpatibility. 

I'd love to hear your thoughts about this :)

ThibaultLemaire:
Closes #233

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ setup(
     url='https://github.com/touilleMan/umongo',
     packages=['umongo', 'umongo.frameworks'],
     include_package_data=True,
+    package_data={'umongo': ["py.typed"]},
     python_requires='>=3.7',
     install_requires=requirements,
     extras_require={


ThibaultLemaire:
For the record, running mypy on the current codebase yields 44 errors which I am not going to list here, though I will note that most of them are caused by missing annotations on dependencies (e.g. `error: Cannot find implementation or library stub for module named "pymongo"`).

Here is the list:

```
pymongo
pymongo.errors
pymongo.database
pymongo.cursor
bson
motor.motor_asyncio
marshmallow
twisted.internet.defer
txmongo
txmongo.database
mongomock.database
mongomock.collection
```

`pymongo` and `bson` might be fixed with [`pymongo-stubs`](https://pypi.org/project/pymongo-stubs/) but I do not have a solution for the rest which is an other reason why I prefer to go with stubs for umongo

ThibaultLemaire:
Ok, so, now that I have a more complete set of stubs I'm hitting what @talwrii warned about. That is: metaprogramming.

E.g.: My previous example was incorrect: without registering my `Document` to an instance I shouldn't be able to call `find` on it.

Here is an update example (note the `@instance.register` decorator):

```python
from umongo import Document
from umongo.fields import StrField
from umongo.frameworks import MotorAsyncIOInstance

instance = MotorAsyncIOInstance()

@instance.register  # Changes the parent type to MotorAsyncIODocument
class MyDocument(Document):
  known_field = StrField(required=True)

    
MyDocument.find()  # With motor, this isn't a cursor, but a coroutine that returns a cursor, and this is where mypy can remind us to await it.
```

Naively, this shouldn't be too hard to type correctly. A decorator boils down to a function that takes a function/class definition and returns an other function/class definition. So typing the decorator properly should be enough for mypy to work its magic, right? This is what I tried:

```python
# umongo/frameworks/motor_asyncio.pyi
...

class MotorAsyncIOInstance(Instance): 
    ...
    def register(self, template: Type[DocumentTemplate]) -> Type[MotorAsyncIODocument]: ...
```

Except that
> class decorators are handled differently than function decorators in mypy: decorating a class does not erase its type

and here is what I get from mypy when checking the previous code:

```
main:12: error: "Type[MyDocument]" has no attribute "find"  [attr-defined]
```

Which is true for a `umongo.document.DocumentTemplate` but not for a `umongo.frameworks.motor_asyncio.MotorAsyncIODocument` which is what it actually is at runtime...

I suppose a mypy plugin could work that out and fix the situation, but that feels a bit dirty.

ThibaultLemaire:
I originally planned to write a mypy plugin that could lint pymongo queries on several ODMs (not only `umongo` but also `mongoengine`, or `pydantic-odm`) but maybe that was a bit ambitious, and I am short on time to work on all of this. (This is why I needed this PR initially)

This is what it is [currently](165vcxd8514pkh42k8x2sz4mkr4j11v17lng2k6pk7rrcrw9mkyr) able to do: 

```python
from umongo import Document
from umongo.fields import StrField
from umongo.frameworks import PyMongoInstance

instance = PyMongoInstance()

@instance.register
class MyDocument(Document):
  known_field = StrField(required=True)

    
cursor = MyDocument.find({"known_field": 123, "unknown_field": "string"}) # E: main.MyDocument has no field "unknown_field"  [misc]
reveal_type(cursor) # N: Revealed type is "umongo.frameworks.pymongo.WrappedCursor[main.MyDocument*]"
```

Note that with only the type stubs I wrote in this PR (so, without the plugin), this is what mypy will yield:

```python
from umongo import Document
from umongo.fields import StrField
from umongo.frameworks import PyMongoInstance

instance = PyMongoInstance()

@instance.register
class MyDocument(Document):
  known_field = StrField(required=True)

    
cursor = MyDocument.find({"known_field": 123, "unknown_field": "string"}) # E: "Type[MyDocument]" has no attribute "find"  [attr-defined]
reveal_type(cursor) # N: Revealed type is "Any"
```

That's because of a limitation of mypy in how it handles class decorators. **So this PR is effectively useless without this plugin...**

So because this plugin has somewhat shifted in scope towards being more `umongo`-specific, I don't believe it makes sense for it to live in a different repository anymore.

I'm thinking I should move it over here and merge it to this PR, maybe in a folder called `mypy` like what pydantic does? What do you think @lafrech ?

ThibaultLemaire:
I have since changed jobs and this is no longer relevant to my current job, so I'm sorry but I won't be working on it anymore (however anyone is welcome to pick up where I left off, and ask for information if there's something I overlooked).

If you are looking for a Pydantic<->Mongo ODM I think you should look into [`pydantic-odm`](https://github.com/i8enn/pydantic-odm) if you're OK with handling indexes yourself. It's a bit lacking in documentation but IIRC the API is straight-forward enough that you can figure it out for yourself by poking around. (uMongo uses Marshmallow internally which isn't ideal if in the end you want Pydantic. You'd have to go through BSON->Marshmallow->Pydantic and back again.)

msenol86:
@ThibaultLemaire I think [this](https://github.com/roman-right/beanie/) ODM is better than pydantic-odm, what do you think?

ThibaultLemaire:
Indeed. I wish it had existed back when I was making my own search.

PR #364, opened 2021-10-15
Trigger set_modified for insert on the ListField

denya:
The problem is that if you have a `ListField` and you do `my_field.inset(0, value)` it won't mark the field as `modified` and won't update it on `commit()`.


denya:
Hey, @lafrech

Please, check the new commit, I've added some new code to the `text_complex_list`

Issue #365, opened 2021-12-18
pymongo Database class explicitly denies bool


whophil:
In pymongo >= 4, the `Database` class denies usage as bool. This was introduced with 
https://github.com/mongodb/mongo-python-driver/pull/728

This breaks https://github.com/Scille/umongo/blob/28168d67f1c0a5030b65da1717859bd7c3a0bba0/umongo/instance.py#L136, and perhaps other lines.

Workaround is to use pymongo < 4 until fixed here.

lafrech:
True. Would you like to send a fix PR?

whophil:
> True. Would you like to send a fix PR?

Will do, when I get a moment. But I won't be offended if somebody beats me to it 🙂

lafrech:
Closed by #366.

Thank you for your contributions.

I just sent you an invitation to join the maintenance team. No pressure to do any more work but if you have things you'd like to improve, here are the keys.

PR #366, opened 2021-12-22
Fix issue where pymongo Database raises if bool'd

whophil:
Addresses #365 

Now it just checks if the `db` is `None`.

Perhaps there are some edge cases with other database types which _do_ implement `bool`?

If so, we can move this change down to the `PyMongoInstance` subclass, but it would require a little more rework.

### DIFF
--- a/umongo/document.py
+++ b/umongo/document.py
@@ -117,7 +117,7 @@ class MetaDocumentImplementation(MetaImplementation):
         """
         if cls.opts.abstract:
             raise NoDBDefinedError('Abstract document has no collection')
-        if not cls.opts.instance.db:
+        if cls.opts.instance.db is None:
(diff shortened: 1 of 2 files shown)


lafrech:
Good, thanks.

Good point about other DB types. Can't investigate right now. Do you think you could have a look at it?

Thanks.

whophil:
In a quick check, none of `motor`, `mongomock`, or `txmongo` Database types implement `__bool__`, so changing the comparison to `is None` should be equivalent.

- motor: https://github.com/mongodb/motor/blob/master/motor/core.py#L491
- mongomock: https://github.com/mongomock/mongomock/blob/develop/mongomock/database.py#L37
- txmongo: https://github.com/twisted/txmongo/blob/master/txmongo/database.py#L13

PR #367, opened 2021-12-22
Add fields to EmbeddedDocument.__dir__() to facilitate dynamic code completion

whophil:
I may be misremembering, but I thought this worked at some point in the past.

### DIFF
--- a/tests/test_document.py
+++ b/tests/test_document.py
@@ -126,6 +126,13 @@ class TestDocument(BaseTest):
         with pytest.raises(KeyError):
             del john['missing']
 
+    def test_dir(self):
+        john = self.Student.build_from_mongo(data={
+            'name': 'John Doe', 'birthday': dt.datetime(1995, 12, 12), 'gpa': 3.0})
+        assert 'name' in dir(john)
+        assert 'birthday' in dir(john)
+        assert 'gpa' in dir(john)
+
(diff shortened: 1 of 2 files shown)


whophil:
Failing due to issues addressed by #366 , rerun tests after that is merged.

lafrech:
Thanks for contributing.

I'm not familiar with this. I've read about `dir` today and it looks like overriding `__dir__` is not so common.

Anyway, I have no objection. Should the same be done with `Document`?

whophil:
> I'm not familiar with this. I've read about `dir` today and it looks like overriding `__dir__` is not so common.

To provide some more context - this aids in dynamic auto-completion, so that in IDEs or REPLs that support it (e.g., IPython), dot completion will show you the field names as suggestions, _in addition to_ all the suggestions for `my_document`.

Before:
image

With `__dir__` override
image

> Anyway, I have no objection. Should the same be done with `Document`?

It looks like `Document` already inherits `__dir__` from `EmbeddedDocument`. Also, the test for this capability is actually on `Document` and not `EmbeddedDocument`, so perhaps this is enough as-is.

lafrech:
Oh, right. `Document` inherits it from `EmbeddedDocument`.

Yeah, we could test it on both, but it is not critical.

Issue #196, opened 2019-05-22
Feature request: fetch all documents in Reference FIeld list with projection


halfdanrump:
Hi there! 

Using the example class from the docs:
```
@instance.register
class User(Document):
    email = fields.EmailField(required=True, unique=True)
    birthday = fields.DateTimeField(validate=validate.Range(min=datetime(1900, 1, 1)))
    friends = fields.ListField(fields.ReferenceField("User"))

    class Meta:
        collection = db.user
```
I'd like to be able to do a query like
```
friends = User.find_one().friends
friends.fetch_all(projection=['name'])
```
Does it make sense? 

Thank you :)

halfdanrump:
For the record, I'll be happy to work on this if you think the feature makes sense :)

lafrech:
IIUC, this is two features.

- Add `projection` to `fetch`. Already discussed in https://github.com/Scille/umongo/issues/98. It revealed more complicated than I first thought and I gave it up. I still think it would be a nice feature. For this feature, let's continue discussion there.

- Add `fetch_all` method to `List`. Awkward because a `ListField` can contain any field, not necessarily `EmbeddedField`.

    With `projection` in `fetch`, you could do
    ```py
    fetched_friends = [f.fetch(projection=['name']) for f in friends]
    ```
    Not as straightforward but not that bad.

halfdanrump:
You're right, these are two separate features

> With projection in fetch, you could do
> 
> fetched_friends = [f.fetch(projection=['name']) for f in friends]

Won't this send `n` queries to mongo where `n` is the number of friends?

Leaving aside the projection for now, how about a solution where 
```
User.find_one().friends.fetch_all()
```
works if and only if the entries in the `ListField` is a `ReferenceField`?

Issue #98, opened 2017-03-07
Add projection feature to Reference.fetch()


lafrech:
As proposed in [this comment](https://github.com/Scille/umongo/pull/96#issuecomment-284217501).

I would try to have fetch signature close to find/find_one signatures, for consistency. Something like this, so we could pass projection or any other `find_one` argument:

```python
    def fetch(self, *args, force_reload=False, **kwargs):
        if not self._document or force_reload:
            if self.pk is None:
                raise ReferenceError('Cannot retrieve a None Reference')
            self._document = self.document_cls.find_one(self.pk, *args, **kwargs)
            if not self._document:
                raise ValidationError(self.error_messages['not_found'].format(
                    document=self.document_cls.__name__))
        return self._document
```
There's a trap, here, though, due to the caching mechanism.

Just like you can load partial documents, and when you do that you should be aware that some fields are not loaded, you could partially fetch and when doing so you should be aware of the partial load when accessing fetched document.

The trick is that validation of reference fields when saving referencer does a fetch() as well, and for performance reasons, I'd really like to use a projection here, but this populates the cache behind our back with an almost empty document.

This would not happen if we did not use fetch in `io_validate`:

```python
def _reference_io_validate(field, value):
    if value.document_cls.find_one(value.pk, projection={'_id': True}) is None:
        raise ValidationError(value.error_messages['not_found'].format(
            document=value.document_cls.__name__))
```

In fact, this would make validation more efficient, I suppose, because we would not instantiate every referenced Document when saving referencer.

This code is pretty much what I suggested to put in an `exists` method of Reference. You said you didn't like the idea ("too low level").

```python
class PyMongoReference(Reference):

    @property
    def exists(self):
        return self.document_cls.find_one(self.pk, projection={'_id': True}) is not None

def _reference_io_validate(field, value):
    if not value.exists:
        raise ValidationError(value.error_messages['not_found'].format(
            document=value.document_cls.__name__))
```

I thing I'm beginning to see what you meant by "too low level". Should we have some sort of `cook_find_projection` function in `tools.py`? AFAIU, we can currently pass a `projection` argument to `find_one` but we should feed it DB names, not OO names, right?

touilleMan:
> This code is pretty much what I suggested to put in an exists method of Reference. You said you didn't like the idea ("too low level").

Well you're proving me there is a valid usecase ;-)
[...]

PR #380, opened 2022-09-16
Add projection argument to Reference.fetch()

whophil:
Adds projection argument to `Reference.fetch()`, allowing partial reference documents to be returned.

Addresses _half_ of issue #196

### DIFF
--- a/umongo/data_objects.py
+++ b/umongo/data_objects.py
@@ -178,12 +178,11 @@ class Reference:
             a projection which limits the data returned from database.
         """
         raise NotImplementedError
-    # TODO replace no_data by `exists` function
 
     @property
     def exists(self):
         """
-        Check if the reference document exists in its collection.
+        Check if the reference document exists in the database.
         """
         raise NotImplementedError
 
--- a/umongo/data_objects.py
+++ b/umongo/data_objects.py
@@ -180,6 +180,13 @@ class Reference:
         raise NotImplementedError
     # TODO replace no_data by `exists` function
 
+    @property
+    def exists(self):
+        """
+        Check if the reference document exists in its collection.
+        """
+        raise NotImplementedError
+
     def __repr__(self):
         return '<object %s.%s(document=%s, pk=%r)>' % (
             self.__module__, self.__class__.__name__, self.document_cls.__name__, self.pk)
--- a/umongo/data_objects.py
+++ b/umongo/data_objects.py
@@ -169,13 +169,13 @@ class Reference:
         """
         Retrieve from the database the referenced document
 
-        :param no_data: if True, the caller is only interested in whether or
-            not the document is present in database. This means the
+        :param no_data: if True, the caller is only interested in whether
+            the document is present in database. This means the
             implementation may not retrieve document's data to save bandwidth.
         :param force_reload: if True, ignore any cached data and reload referenced
             document from database.
-        :param projection: if supplied, this is a dictionary describing a projection
-            which limits the data returned from database.
+        :param projection: if supplied, this is a dictionary or list describing
+            a projection which limits the data returned from database.
         """
         raise NotImplementedError
     # TODO replace no_data by `exists` function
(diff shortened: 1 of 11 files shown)


umongo/frameworks/pymongo.py
```diff
@@ -322,11 +322,14 @@ def __init__(self, *args, **kwargs):
...
-    def fetch(self, no_data=False, force_reload=False):
+    def fetch(self, no_data=False, force_reload=False, projection=None):
         if not self._document or force_reload:
             if self.pk is None:
                 raise NoneReferenceError('Cannot retrieve a None Reference')
-            self._document = self.document_cls.find_one(self.pk)
+            if projection is None:
+                self._document = self.document_cls.find_one(self.pk)
+            else:
+                self._document = self.document_cls.find_one(self.pk, projection)
```
lafrech, line 332: Why not a single case

    self._document = self.document_cls.find_one(self.pk, projection=projection)

Doesn't this work for all cases?

lafrech (commented):
I left a minor comment.

The issue you point to points to #98 which raises concerns about projection and caching mechanism. And the fact that DB names should be expected (unless umongo tries to translates field names into DB names but I think it wouldn't be the only place where DB names are needed).

In any case, you may merge this if you're happy with it. If we can't have a perfect solution, this is better than nothing. It would be nice if known limitations were documented, though.

whophil:
Thanks @lafrech. I had actually not seen the discussion on issue #98. Let me think over this a bit and potentially rework.

whophil:
@lafrech I've implemented `cook_find_projection`, as suggested in #98, so that the projections can use uMongo field names and not DB names.

Other things discussed in #98 which I did not address:
- The cached `_document` is just the first-fetched document, not a cache which is looked up by the projection. The projection-keyed cache may still be useful in the future. If implemented, it should have a configurable max size.
- `_reference_io_validate` is untouched (it still uses `.fetch()`).

whophil:
FYI, the suggestion to not use`.fetch()` in `_reference_io_validate` makes a huge difference in performance in some cases.

In my application, I have a document which references over a thousand others. The time to `.commit()` the parent document went from 21.8 seconds to 1.5 seconds (PyMongo with mongo running on localhost).

umongo/data_objects.py
```diff
@@ -165,19 +165,28 @@ def __init__(self, document_cls, pk):
...
+        :param no_data: if True, the caller is only interested in whether
+            the document is present in database. This means the
             implementation may not retrieve document's data to save bandwidth.
         :param force_reload: if True, ignore any cached data and reload referenced
             document from database.
+        :param projection: if supplied, this is a dictionary or list describing
+            a projection which limits the data returned from database.
         """
         raise NotImplementedError
     # TODO replace no_data by `exists` function
```
lafrech, line 181: I guess we could deprecate `no_data`, then? (No pressure to do this in this PR.)

lafrech (approved):
LGTM!

You may merge and release!

umongo/data_objects.py
whophil: Sounds good, this can be deprecated in a future PR/release.

whophil:
@lafrech how do I make the CI "go" on this?

lafrech:
That's a good question. I see no obvious reason. Hooks are enabled. Can't look into this right now.

Frankly, Azure CI is a nightmare to configure and rather than trying to fix this I'd be tempted to move to GitHub actions. You may give it a try in another PR if you want. Setting up tests is relatively straightforward. See what I did on flask-smorest. The tricky part would be launching MongoDB.

------------------------------

OK I peeked into Azure logs and CI did go but apparently didn't comment here with the results. Note sure if the pipeline logs there are public.

Here's the last result:

```
./tests/conftest.py:38:101: E501 line too long (106 > 100 characters)
./umongo/frameworks/motor_asyncio.py:415:101: E501 line too long (105 > 100 characters)
```

whophil:
> OK I peeked into Azure logs and CI did go but apparently didn't comment here with the results. Note sure if the pipeline logs there are public.

I don't know where to find the link to these results, so I can't tell you if they're public or not :)

I fixed the too-long lines. Merging now!

As far as a release goes - what is the process on this project?

PR #387, opened 2024-04-30
deps: Upgrade `motor` and add pymongo 4.x to test matrix

### LINKED ISSUE #386: [dependencies] Upgrade `pymongo>4` and `motor>3`.
Hi @lafrech 👋 

Thank you for maintaining `umongo`, a product we maintain depends on it! 🙇 

Lately it's come to our attention that we're not using the latest `pymongo` version and this is due to `umongo[motor]` being pinned to `motor>2,<3`.

Are you aware of a specific reason why motor is pinned `<3`? 
It started supporting `pymongo>4` after 3.0 unfortunately.

If it helps I can open a PR to investigate motor 3.x support in umongo 🙌

alexandervaneck:
Fixes #386 and #278


setup.py
```diff
@@ -32,7 +32,7 @@
     python_requires='>=3.7',
     install_requires=requirements,
     extras_require={
-        'motor': ['motor>=2.0,<3.0'],
+        'motor': ['motor>=2.0,<4.0'],
```
alexandervaneck, line 35: @lafrech should we unpin this instead?

lafrech: No, please don't unpin. This is fine as is it.

lafrech:
I think the second link in the OP is wrong (278 is an unrelated PR).

Not sure why Azure CI don't run. In my other projects, we have switched to GH actions. Configuring Azure CI is a pain.

It would be nice to update this repo to match the CI settings from marshmallow (pyproject.toml, ruff, GH actions). If you want to take a stab at it, you're very welcome. But don't feel like you have to. However, rather than struggling with Azure, I suggest you don't wait for me to carve some time to do the GH action move and you try to contribute that change. You may want to do that in another PR for clarity. marshmallow is a nice example to get inspiration from.

alexandervaneck:
> I think the second link in the OP is wrong (278 is an unrelated PR).

You were right - I confused a ticket on txmongo. I've removed the link.

alexandervaneck:
> Not sure why Azure CI don't run. In my other projects, we have switched to GH actions. Configuring Azure CI is a pain.
> 
> It would be nice to update this repo to match the CI settings from marshmallow (pyproject.toml, ruff, GH actions). If you want to take a stab at it, you're very welcome. But don't feel like you have to. However, rather than struggling with Azure, I suggest you don't wait for me to carve some time to do the GH action move and you try to contribute that change. You may want to do that in another PR for clarity. marshmallow is a nice example to get inspiration from.


Thank you for clarifying. I unfortunately don't have the time to fixup the build process. 🙏 So I suppose this will have to wait until you have time.

ggdupont:
Any chance to see this PR merged soon? or should we fork?

lafrech:
I don't use umongo anymore but I'm happy to merge PRs. People seriously tempted to fork are very welcome to join in.

Ideally I'd fix/update CI before merging. Anyone willing to copy the CI config from marshmallow, that would be fantastic.

We could merge this change before that but the CI needs to be fixed at some point anyway.

PR #391, opened 2025-09-19
Upgrade umongo for Python 3.12 compatibility

sanjusci:
feature: upgrade umongo for Python 3.12 compatibility

Replaced deprecated marshmallow missing with load_default / dump_default
Updated asyncio usage:
Replaced asyncio.coroutine with asyncio.iscoroutinefunction + async wrappers
Migrated validator runner to asyncio.gather
Fixed txmongo dict_items TypeError by casting to list before sort
Updated MotorAsyncIO integration:
Conditional import of motor with pytest skip markers if missing
Adjusted test cases for default handling and async validation
Added compatibility shims for smooth migration from Python <3.12


.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
lafrech, line 1: The whole .idea directory should not be added to the repo.

This is IDE specific. You should use a global .gitignore to ignore that.

=0.21.0
```diff
@@ -0,0 +1,6 @@
+Requirement already satisfied: pytest-asyncio in /home/global/miniconda3/envs/umongo/lib/python3.12/site-packages (1.2.0)
```
lafrech, line 1: Not sure what those two files are but I don't think they belong here either.

setup.py
```diff
@@ -17,11 +17,12 @@
 requirements = [
     "marshmallow>=3.10.0",
     "pymongo>=3.7.0",
+    "motor>=3.1.1",
```
lafrech, line 20: This shouldn't be a requirement. Let's keep it as an extra below.

setup.py
```diff
@@ -32,9 +33,10 @@
     python_requires='>=3.7',
     install_requires=requirements,
     extras_require={
-        'motor': ['motor>=2.0,<3.0'],
+        'motor': ['motor>=3.1.1'],
         'txmongo': ['txmongo>=19.2.0'],
         'mongomock': ['mongomock'],
+        'marshmallow': ['marshmallow>=3.14.0']
```
lafrech, line 39: Bump marshmallow requirement above, don't make it an extra.

lafrech (changes requested):
Thanks. I didn't inspect every detail but this looks good.

Unfortunately, the CI doesn't run (see #387). I don't use umongo anymore so I doubt I'll get the time to tackle this. Copying marshmallow CI GH actions would be nice. I can take a little time to set required secret token if needed.

.idea/.gitignore
sanjusci: Remove from code
Thanks

setup.py
sanjusci: Done

lafrech (changes requested):
Looks good. Thanks again for tackling this.

Please cherry-pick the last commit from the [test-ci branch](https://github.com/Scille/umongo/tree/ci-test) I just pushed, so that CI is executed using GH actions. Tests should pass.

After that, we shall remove support for outdated Python versions (3.7 and 3.8).

If needed, we can also drop old dependencies versions if some of them are obsolete, and make this a major release. Are there dependencies versions you think we could drop (e.g. motor 2, pymongo 3)? I haven't been using Mongo DB for ages, so I have no idea.

I still need to setup PyPI. I'll do that when needed.

lafrech:
@sanjusci I just published 4.0.0b1. Please test and provide feedback.

See #404.

